PR #9, opened 2018-04-02
Unify network id for development

fsmanuel:


bumi:
in the future we will need remove the development contract addresses from the repo/npm package.

PR #11, opened 2018-04-04
Add a general purpose contract CLI

bumi:
This allows to call any method on any contract using the CLI.

for example:

  $ truffle exec scripts/cli.js Operator proposalsCount


bumi:
yep, should have a linter here I guess. my JS style is not very good I guess - need to teach my editor some tricks.

PR #20, opened 2018-04-08
[WIP] Add function to batch votings

bumi:
The batchVote function accepts an array of proposal ids and votes for
every one.
Normally arrays without fix length are problematic and gas usage can not be
estimated really well. So we need to see how that works or what other
pattern could be used.

### DIFF
--- a/contracts/Operator.sol
+++ b/contracts/Operator.sol
@@ -95,6 +95,12 @@ contract Operator is Upgradeable {
     ProposalCreated(proposalId, msg.sender, p.recipientId, p.amount);
   }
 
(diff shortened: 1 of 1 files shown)


skddc:
lgtm :+1:

bumi:
generally we need to keep an eye on the fees as those will ne n times vote(). and maybe n proposals will be executed (also see #10 for discussion)

skddc:
> generally we need to keep an eye on the fees as those will ne n times vote(). and maybe n proposals will be executed (also see #10 for discussion)

That would be an acceptable scenario, as one would have to pay the same fee for all seperate vote transactions in this case.

PR #24, opened 2018-04-10
More consistent contract parameter naming

bumi:
because we all love consistency. 

Please comment and mark lines that you think should have different/more descriptive/consistent naming

### DIFF
--- a/contracts/Operator.sol
+++ b/contracts/Operator.sol
@@ -24,9 +24,9 @@ contract Operator is Upgradeable {
   mapping(uint256 => Proposal) public proposals;
   uint256 public proposalsCount;
 
-  event ProposalCreated(uint256 id, address creator, uint recipient, uint256 amount);
+  event ProposalCreated(uint256 id, address creator, uint recipientId, uint256 amount);
   event ProposalVoted(uint256 id, address voter, uint256 totalVotes);
-  event ProposalExecuted(uint256 id, uint recipient, uint256 amount);
+  event ProposalExecuted(uint256 id, uint recipientId, uint256 amount);
 
   modifier coreOnly() { 
     require(contributorsContract().addressIsCore(msg.sender));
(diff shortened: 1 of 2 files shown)


contracts/Operator.sol
```diff
@@ -24,9 +24,9 @@ contract Operator is Upgradeable {
   mapping(uint256 => Proposal) public proposals;
   uint256 public proposalsCount;
 
-  event ProposalCreated(uint256 id, address creator, uint recipient, uint256 amount);
-  event ProposalVoted(uint256 id, address voter, uint256 totalVotes);
-  event ProposalExecuted(uint256 id, uint recipient, uint256 amount);
+  event ProposalCreated(uint256 id, address creatorAccount, uint256 recipientId, uint256 amount);
```
fsmanuel: What about `recipientId` -> `contributorId`?

contracts/Operator.sol
```diff
@@ -109,33 +94,29 @@ contract Operator is Upgradeable {
...
-  function vote(uint256 _proposalId) public coreOnly returns (uint _pId, bool _executed) {
-    var p = proposals[_proposalId];
+  function vote(uint256 proposalId) public coreOnly {
+    var p = proposals[proposalId];
     require(!p.executed);
-    uint256 contributorId = contributorsContract().getContributorIdByAddress(msg.sender);
-    require(p.votes[contributorId] != true);
-    p.voterIds.push(contributorId);
-    p.votes[contributorId] = true;
+    uint256 voterId = contributorsContract().getContributorIdByAddress(msg.sender);
```
fsmanuel: Why the change from `contributorId` -> `voterId`?

bumi: yeah...hmmm.. thought it is more descriptive.. but I see your point.

contracts/Operator.sol
bumi: In these events we have two different "contributors" 
one is the recipient and one is the voter - that's why I'd use voterId and recipientId. 
?

PR #26, opened 2018-04-15
Add batch voting for proposals

bumi:
The batchVote function accepts an array of proposal ids and votes for
every one.
Normally arrays without fix length are problematic and gas usage can not be
estimated really well. So we need to see how that works or what other
pattern could be used.


fsmanuel:
@bumi can you rebase? I'm working on `kredits-web` for this. Need to test if it works.

bumi:
done, conflict should be resolved.

fsmanuel:
I need to build the abis again, right?

bumi:
yes, otherwise the JS does not know about the new function. 
`npm run build-json` should do the trick? 

I am not sure if it is good to add the ABI changes in the PR it could easily cause conflicts - similar to a gemfile.lock or similar. - So I think it is better to build them in master? what do you think? how should we handle this in the future?

fsmanuel:
I was wondering the same. I guess if we rebase/merge master before we merge the PR we should commit the ABI changes in the PR.

PR #32, opened 2018-04-19
Use prompt instead of argv arguments in scripts

bumi:
This makes it easier to handle truffle arguments which we for example need to specify the network.
So we ask the user for input instead on using the argv array which might change beause of the truffle arguments used.


fsmanuel:
Closing in favor of #33 and #35

PR #36, opened 2018-04-20
Refactor helper scripts to use kredits module

bumi:
This reuses the kredits library functions in the helper scripts and seeds. 
We no longer need to add IPFS hashes manually but simply can provider contributor/proposal data.

depends on #34


package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "build-json": "truffle compile && node ./scripts/build-json.js",
-    "bootstrap": "truffle migrate --reset && truffle exec scripts/seeds.js && npm run build-json",
+    "bootstrap": "truffle migrate --reset && npm run build-json && truffle exec scripts/seeds.js",
```
fsmanuel: 👍

Issue #37, opened 2018-04-21
Make IPFS config configurable


fsmanuel:
We need to make IPFS config configurable for the use with `.init()`.
Right now we use the default that works for development but not if we deploy.

- [ ] Set it in `kredits-web`

bumi:
done in: #40

PR #40, opened 2018-04-23
Constructor configuration options

bumi:
This changes the function signature of the constructor but allows us to
pass any options.
Mainly this is now used for the ipfs configs.

Ref #37

### DIFF
--- a/lib/kredits.js
+++ b/lib/kredits.js
@@ -22,16 +22,16 @@ function capitalize(word) {
 
 class Kredits {
 
-  constructor(provider, signer, addresses) {
+  constructor(provider, signer, options = {}) {
     this.provider = provider;
     this.signer = signer;
 
     // by default we only need the registry address.
     // the rest is loaded from there in the init() function
-    this.addresses = addresses || {Registry: RegistryAddress[this.provider.chainId.toString()]}; // chaiID must be a string
-    this.abis = ABIS;
+    this.addresses = options['addresses'] || { Registry: RegistryAddress[this.provider.chainId.toString()] }; // chaiID must be a string
+    this.abis = options['abis'] || ABIS;
+    this.ipfs = new IPFS(options['ipfsConfig']);
     this.contracts = {};
-    this.ipfs = new IPFS();
   }
 
   init(names) {
@@ -50,11 +50,7 @@ class Kredits {
 
   static setup(provider, signer, ipfsConfig = null) {
     console.log('Kredits.setup() is deprecated use new Kredits().init() instead');
-    let ipfs = new IPFS(ipfsConfig);
-    return new Kredits(provider, signer).init().then((kredits) => {
-      kredits.ipfs = ipfs;
-      return kredits;
-    });
+    return new Kredits(provider, signer, { ipfsConfig: ipfsConfig }).init();
   }
 
   get Registry() {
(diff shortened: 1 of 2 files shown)


lib/kredits.js
```diff
@@ -22,16 +22,16 @@ function capitalize(word) {
 
 class Kredits {
 
-  constructor(provider, signer, addresses) {
+  constructor(provider, signer, options = {}) {
     this.provider = provider;
     this.signer = signer;
```
fsmanuel: What about a destructuring of the options to make it more easy to see the used options?

lib/utils/ipfs.js
```diff
@@ -5,7 +5,7 @@ class IPFS {
 
   constructor(config) {
     if (!config) {
-      config = {host: 'localhost', port: '5001', protocol: 'http'};
+      config = { host: 'localhost', port: '5001', protocol: 'http' };
```
fsmanuel: 👍

lib/kredits.js
```diff
@@ -50,11 +50,7 @@ class Kredits {
 
   static setup(provider, signer, ipfsConfig = null) {
     console.log('Kredits.setup() is deprecated use new Kredits().init() instead');
-    let ipfs = new IPFS(ipfsConfig);
-    return new Kredits(provider, signer).init().then((kredits) => {
-      kredits.ipfs = ipfs;
-      return kredits;
-    });
+    return new Kredits(provider, signer, { ipfsConfig: ipfsConfig }).init();
```
fsmanuel: 👍

lib/kredits.js
```diff
@@ -22,16 +22,16 @@ function capitalize(word) {
...
+  constructor(provider, signer, options = {}) {
     this.provider = provider;
     this.signer = signer;
 
     // by default we only need the registry address.
     // the rest is loaded from there in the init() function
-    this.addresses = addresses || {Registry: RegistryAddress[this.provider.chainId.toString()]}; // chaiID must be a string
-    this.abis = ABIS;
+    this.addresses = options['addresses'] || { Registry: RegistryAddress[this.provider.chainId.toString()] }; // chaiID must be a string
+    this.abis = options['abis'] || ABIS;
```
fsmanuel: You sneaked that one in but I ❤️ it,

skddc: Typo. :)

fsmanuel: iPad...

lib/kredits.js
bumi: something like this: 
` let { address, abis, ipfsConfig } = options ` 
is that correct?

bumi:
what do you mean with " track the update of hubot-kredits in the open issue"?

lib/kredits.js
fsmanuel: Jup

lib/kredits.js
skddc: I meant in the code. Someone coding on an iPad? :D

fsmanuel:
@bumi I don't know why I wrote it... I'll merge.

lib/kredits.js
fsmanuel: Ah you mean the `chaiID` 😆 ? Not yet coding on an iPad...
It was a bad idea to come back to Hamburg. Volley outside...

bumi: :sun_with_face:

PR #41, opened 2018-04-23
Improve helper scripts

bumi:
now using the library in add-proposal.js, too. And the CLI is smarter and allows calling functions on the contract and on the wrapper.

### DIFF
--- a/scripts/add-contributor.js
+++ b/scripts/add-contributor.js
@@ -1,6 +1,5 @@
 const Registry = artifacts.require('./Registry.sol');
 const promptly = require('promptly');
-const bs58 = require('bs58');
 
 const ethers = require('ethers');
 const Kredits = require('../lib/kredits');
@@ -37,7 +36,7 @@ module.exports = function(callback) {
     console.log("\nAdding contributor:");
     console.log(contributorAttributes);
 
-    kredits.Contributor.add(contributorAttributes, {gasLimit: 250000}).then((result) => {
+    kredits.Contributor.add(contributorAttributes, { gasLimit: 250000 }).then((result) => {
       console.log("\n\nResult:");
       console.log(result);
       callback();
--- a/scripts/add-proposal.js
+++ b/scripts/add-proposal.js
@@ -19,7 +19,7 @@ module.exports = function(callback) {
     if (contributor.length < 5) {
       contributorId = contributor;
     } else {
-      contributorId = await contributors.getContributorIdByAddress(contributor);
+      contributorId = await kredits.Contributor.functions.getContributorIdByAddress(contributor);
     }
     console.log(`Creating a proposal for contributor ID #${contributorId}`);
 
--- a/scripts/add-proposal.js
+++ b/scripts/add-proposal.js
@@ -1,38 +1,46 @@
 const Registry = artifacts.require('./Registry.sol');
-const Operator = artifacts.require('./Operator.sol');
-const Contributors = artifacts.require('./Contributors.sol');
 const promptly = require('promptly');
 
-const bs58 = require('bs58');
-
-function getBytes32FromMultiash(multihash) {
-  const decoded = bs58.decode(multihash);
-
-  return {
-    digest: `0x${decoded.slice(2).toString('hex')}`,
-    hashFunction: decoded[0],
-    size: decoded[1],
-  };
-}
+const ethers = require('ethers');
+const Kredits = require('../lib/kredits');
 
 module.exports = function(callback) {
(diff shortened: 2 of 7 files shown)


scripts/add-contributor.js
```diff
@@ -37,7 +36,7 @@ module.exports = function(callback) {
     console.log("\nAdding contributor:");
     console.log(contributorAttributes);
 
-    kredits.Contributor.add(contributorAttributes, {gasLimit: 250000}).then((result) => {
+    kredits.Contributor.add(contributorAttributes, { gasLimit: 250000 }).then((result) => {
```
fsmanuel: Do you have a linter now?

scripts/add-proposal.js
```diff
@@ -1,38 +1,46 @@
...
+    const kredits = await Kredits.setup(provider, provider.getSigner());
+
+    console.log(`Using operator at: ${kredits.Operator.contract.address}`);
+
+    let contributor = await promptly.prompt('Contributor (address or id): ');
+    let contributorId;
+    if (contributor.length < 5) {
+      contributorId = contributor;
+    } else {
+      contributorId = await contributors.getContributorIdByAddress(contributor);
```
fsmanuel: Where does `contributors` come from?

scripts/cli.js
```diff
@@ -7,38 +7,51 @@ const Kredits = require('../lib/kredits');
...
     const provider = new ethers.providers.Web3Provider(
       web3.currentProvider, { chainId: networkId }
     );
     const kredits = await Kredits.setup(provider, provider.getSigner());
 
-    const contract = kredits[contractName].contract;
-    console.log(`Using ${contractName} at ${contract.address}`);
-    console.log(`Calling ${method} with ${JSON.stringify(args)}`);
+    let contractName = await promptly.prompt('Contract Name: ');
+    const contractWrapper = kredits[contractName];
```
fsmanuel: So good that we went with upper case contract names and the properties instead of the functions!
But it will breaks on `Contributor`, wouldn’t it? I think we should rename that contract next and align the module code to it. Will make the future much brighter!

scripts/cli.js
```diff
@@ -7,38 +7,51 @@ const Kredits = require('../lib/kredits');
...
 
-    if (!contract[method]) {
+    let method;
+    method = await promptly.prompt('Function (? for available functions): ');
+    while (method === '?') {
```
fsmanuel: 👀 nice!

scripts/cli.js
```diff
@@ -7,38 +7,51 @@ const Kredits = require('../lib/kredits');
...
+      method = await promptly.prompt('Function: ');
+    }
+    if (!contractWrapper[method] && !contractWrapper.functions[method]) {
       callback(new Error(`Method ${method} is not defined on ${contractName}`));
       return;
     }
+    let argumentInput = await promptly.prompt('Arguments (comma separated): ', { default: '' });
+    let args = [];
+    if (argumentInput !== '') {
+      args = argumentInput.split(',').map(a => a.trim());
```
fsmanuel: Haha master of es6. I think you don’t need the if and the initial `args`
`let args = argumentInput.split(',')` should return an empty array.

scripts/add-contributor.js
bumi: nah, not yet. :(

scripts/cli.js
bumi: oh yeah, true.

bumi: ah noo. it returns an array with an empty string, that was the reason for the if. or what am I missing?

scripts/cli.js
bumi: yes, agree on renaming it! and maybe adding an alias for Contributor <=> Contributors

scripts/cli.js
fsmanuel: Ah ok.

skddc: But if input is already not an empty string, wouldn't it return an array with one argument string?

scripts/cli.js
fsmanuel: I opened #44 for the alias

scripts/cli.js
skddc: I was actually asking about the `trim` inside, but no idea why. It's for removing whitespace, so that's still needed.

bumi: yep, trim() for potential whitespace as for example in `"first, second"`

bumi:
mergeable? or is anything missing?

PR #42, opened 2018-04-23
Add filter and find by account function to contributors

bumi:
This allows to filter and find contributors by the account entries.

For example:

```js

Contributor.filterByAccount({site: 'github.com'}); // returns an array of all contributors with github account
Contributor.filterByAccount({site: 'github.com', username: 'bumi'}); // returns an array with bumi as single entry

Contributor.findByAccount({site: 'github.com', username: 'bumi'}); // returns the matching contributor 
Contributor.findByAccount({site: 'github.com'}); // returns the matching contributor - the first with a github.com account entry
```

### DIFF
--- a/lib/contracts/contributor.js
+++ b/lib/contracts/contributor.js
@@ -49,9 +49,10 @@ class Contributor extends Base {
       return contributors[method]((contributor) => {
         if (!contributor.accounts) { return false; }
         return contributor.accounts.find((account) => {
-          return searchEntries.reduce((accumulator, searchValue) => {
-            return accumulator && account[searchValue[0]] === searchValue[1];
-          }, true);
+          return searchEntries.every((item) => {
+            let [ key, value ] = item;
+            return account[key] === value;
+          });
         });
       });
     });


bumi:
how would we do that? because of the promise that is returned by `all()` wouldn't we always need to also return a promise?

skddc:
I think he meant the promise should be fulfilled with an array, if it's called `filter` instead of `find`, so it's not misleading.

bumi:
it is returning the filter, line 41... no?

skddc:
It looks like it's returning the result of a `find`, which would be the first matching object in the filtered array.

bumi:
the return value of the find is used in the filter callback. if the callback returns a truthy value the contributor is kept otherwise removed. the return value of filter is returned by the promise. 

probably
```js
return find(...) != undefined;
``` 
would be better instead of simply returning the return value of the find method.

skddc:
Why not just have an explicit `findBy` that actually returns a single contributor?

bumi:
because this way you can filter and not only get one single contributor.

skddc:
Yes, but you can have both `filterBy` and `findBy`. Why only have one function to do it all?

bumi:
not sure if I understand what you mean. 
I try to get contributors by their account entry. that's why I filter the contributors array and find in the accounts array for matching entries. 

the function is only doing one thing. 
a `findBy..` would be a `filterBy()[0]` ?

skddc:
> Contributor.filterByAccount(...); // returns bumi

I think both @fsmanuel and I read this as "returns bumi's contributor object", but it actually meant "returns an array with bumi as the only entry". So if it returns the latter, as you explained, then the original comment about it was invalid and just a misunderstanding.

However, what I mean by "why not have a findBy" was: a normal use case in clients is to actually find a single contributor's data. Then, filtering for exact values, implicitly expecting a single-entry array, is a bit clunky.

(I think the "give me all users with GitHub accounts" doesn't currently appear in kredits-web or hubot-kredits, but the "give me the data for @bumi on GitHub" does.)

bumi:
yes I see. the comment in the PR description is missleading. I've updated that. 
The method that this PR adds always returns an array - it returns the return value of the native `Array.filter` function. 

I tried to implement a reusable function to find a single contributor but thought this logic is more flexible and used `filter` instead of `find`. (line 41)

should I change it to `findByAccount` and use `find` instead of `filter`? 
or add a `findByAccont` function that returns the fist value of the `filterByAccount` function? 
or refactor it that it is possible to call filter or find? 

(in the code right now, with this implementation, I'd use `filterByAccount({})[0]` )

skddc:
I'd add a `findBy`, but actually only call find in that one, and call `filter` in the filter one. It's not clear that a `filter` function would include a `find`, thus restricting results to a single match. If we have both, then filter can filter, and find can find. :)

bumi:
what do you think of this now?

lib/contracts/contributor.js
```diff
@@ -34,6 +34,29 @@ class Contributor extends Base {
...
+  }
+
+  _byAccount(search, method = 'filter') {
+    return this.all().then((contributors) => {
+      const searchEntries = Object.entries(search);
+
+      return contributors[method]((contributor) => {
+        if (!contributor.accounts) { return false; }
+        return contributor.accounts.find((account) => {
+          return searchEntries.reduce((accumulator, searchValue) => {
```
fsmanuel: I would replace the `reduce` with an `every` [Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every)

```js
searchEntries.every((item) => {
  const [ key, value ] = item;
  return account[key] === value;
});
```

bumi: oh, reads nice. thanks.

PR #49, opened 2018-06-07
Rename healthcheck to PreflightChecker

bumi:
see related https://github.com/67P/kredits-web/pull/59

### DIFF
--- a/lib/kredits.js
+++ b/lib/kredits.js
@@ -1,7 +1,7 @@
 const ethers = require('ethers');
 const RSVP = require('rsvp');
 
-const PreflightChecker = require('./utils/preflight_checker');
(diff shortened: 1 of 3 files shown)


lib/kredits.js
```diff
@@ -1,7 +1,7 @@
 const ethers = require('ethers');
 const RSVP = require('rsvp');
 
-const Healthcheck = require('./utils/healthcheck');
+const PreflightChecker = require('./utils/preflight_checker');
```
skddc: Why not just `Preflight.check()` if there's a checker method in that module anyway? I think it would read/look a bit nicer.

skddc:
Noice!

PR #62, opened 2019-03-24
Build on aragonOS

### LINKED ISSUE #52: Contribution tokens (ERC721)
From IRC:

```
<raucao> we'll use ERC721 tokens which can store metadata (the ipfs hash) and have an id
<raucao> and add a property to those for how many kredits were issued
<raucao> then we can use those both in contracts as well as for history views etc
<raucao> and the normal tokens can be sent around and whatnot
<raucao> then, when we distribute money, we can do that to the normal kredits tokens
<raucao> so a contributor can decide to either get dividends with those, or sell them to someone else
<raucao> and also, we can sell pure kredits tokens that don't represent contributions to people who want to support opensource projects, without them being able to vote with those or anything
<raucao> the way we see that now, it would solve our remaining problems with the current tokens and be super clean as well
```

### LINKED ISSUE #57: update ethers.js to 4.0
Ethers.js released a new version. The new version has some breaking changes but updating should be easy. 
have a look a the [migration guide](https://docs.ethers.io/ethers.js/html/migration.html)

ethers.js is also a dependency in kredits-web and must be updated at the same time.

bumi:
This PR builds on #55 (includes the contribution contract) and refactors the smart contracts to use aragonOS. 

Sadly it got a bit of a big refactoring :man_shrugging: that also follows the typical directory structure for aragonOS based contracts. Those are called "apps" and organized in independent folders (which could also be independent npm packages)

### Upgradeablity 
This allows use to develop and deploy each contract independently and use the aragon upgrading system to upgrade the DAO/our kredits instance that uses the app. 
We don't have to develop our own upgrading proxy and keep our own registry - we can focus on our core contracts/features and less on base infrastructure.

### Permission system
Aragon brings a powerful permission system that allows to evaluate the ACL on runtime and thus to more loosly couply our contracts. 
For example to only allow contributors to vote: The proposal contract does no longer need to know about contributors it simply defines a permission and role. The permission is then configured in the ACL and can later be easly updated without the requirement to change the contract. 
 
 
### `apps/contribution`
manages contributions. currently allows to add and claim contributions. 
missing: veto of contributions

### `apps/contributors`
manages all contributor profiles

### `apps/proposals`
allows to add and vote on proposals that issue contribtions.

### `apps/token`
the actual ERC20 kredits token. 
issued by claiming contributions
 
## Usage of the kredits-contracts wrapper
The usage of the wrapper does not change. It slightly got updated to use the aragon DAO system instead of our own registry.

## About AragonOS
It is important to mention that using aragon tools does NOT mean we run within the aragon app/network! So far we only use the tools provided by aragonOS. These are mainly the permission system and the upgradeability of contracts. 
We can allow our contracts/apps to be used with other aragon "apps" or even within the aragon UI, but that is not planned right now. 
### Links: 
* https://hack.aragon.org/docs/aragonos-intro.html
* https://wiki.kosmos.org/Aragon

closes #57 
closed #52


scripts/deploy-kit.js
```diff
@@ -1,49 +1,39 @@
...
 const libPath = path.join(__dirname, '..', 'lib');
 const addressesPath = path.join(libPath, 'addresses');
 
+const KreditsKit = artifacts.require('KreditsKit')
+
 const ensAddr = process.env.ENS
+const daoFactoryAddress = process.env.DAO_FACTORY
 
-module.exports = async (callback) => {
+module.exports = async function(callback) {
```
fsmanuel: Are you a javascript pro now?

scripts/deploy-kit.js
```diff
@@ -1,49 +1,39 @@
...
+  } else {
+    daoFactory = (await deployDAOFactory(null, { artifacts, verbose: false })).daoFactory
+  }
+  console.log(`Using DAOFactory at: ${daoFactory.address}`)
+
+  let kreditsKit = await KreditsKit.new(daoFactory.address, ensAddr)
+
+  const networkId = parseInt(web3.version.network);
+
+  let addresseFile = path.join(addressesPath, `KreditsKit.json`);
```
fsmanuel: What about:

```js
const KreditsKit = artifacts.require('KreditsKit')
const kreditsKitFile = 'KreditsKit.json'
// ...
let addresseFile = path.join(addressesPath, kreditsKitFile);
```

scripts/new-dao.js
```diff
@@ -0,0 +1,44 @@
...
+    console.log(ret.logs);
+    const installedEvents = ret.logs.filter(log => log.event === 'InstalledApp').map(log => log.args)
+    const deployEvents = ret.logs.filter(log => log.event === 'DeployInstance').map(log => log.args)
+
+    if (deployEvents.length > 1) {
+      callback(new Error("More than one DAO was deployed. Something is wrong"))
+    }
+    const daoAddress = deployEvents[0].dao;
+
+    let addresseFile = path.join(addressesPath, `dao.json`);
```
fsmanuel: Haha, make a function to update the `addresseFile`.

lib/kredits.js
```diff
@@ -54,26 +61,30 @@ class Kredits {
...
+    return this.contractFor('Kernel');
   }
 
   get Contributor() {
     // TODO: rename to contributor
-    return this.contractFor('contributors');
+    return this.contractFor('Contributor');
   }
 
   get Contributors() {
```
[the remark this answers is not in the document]
bumi: not sure, maybe kredits-web is still using it. 
even though this PR changes a lot the API of the wrapper is still the same and I tried to minimize changes there.

scripts/deploy-kit.js
bumi: a pro in being desperate :) 
those scripts are executed in some truffle context and debugging sometimes is a pain because truffle swallows the errors and traces... so I sometimes threw everything on it :D

scripts/new-dao.js
bumi: yep, did that now finally. and added a `helpers' directory... 
do you have a good name for that file/function that injects that data based on the networkId?

bumi:
I've deployed to rinkeby and the local devchain deployments work for me. 
So would be awesome if you could test it and see if you can run it locally on your machine (see readme file) 

To deploy to a testnet or mainnet it now actually also supports [frame](https://frame.sh/) to handle keys, but there is an open [PR](https://github.com/floating/ethereum-provider/pull/3) to make this work with truffle 4 and web3 0.x which we are using.

bumi:
it's been quite a lot of hacking and trial and error to get this running and I need to clean up some code, but feedback is welcome. 
I feel like this is now finally a working smart contract setup to iterate on... let's see if it really works.

contracts/KreditsKit.sol
```diff
@@ -3,27 +3,26 @@ pragma solidity 0.4.24;
...
 
-contract KreditsKit is KitBase, APMNamehashOpen, ACLSyntaxSugar  {
-    bytes32 public contributorAppId = apmNamehash("contributor"); // 0xe9140f1e39c8a1d04167c3b710688a3eecea2976f34735c8eb98956f4764635b
-    bytes32 public contributionAppId = apmNamehash("contribution"); // 0x7fcf91283b719b30c2fa954ff0da021e1b91aed09d7aa13df5e8078a4a1007eb 
-    bytes32 public tokenAppId = apmNamehash("token"); // 0xe04a882e7a6adf5603207d545ea49aec17e6b936c4d9eae3d74dbe482264991a 
-    bytes32 public proposalAppId = apmNamehash("proposal"); // 0xaf5fe5c3b0d9581ee88974bbc8699e6fa71efd1b321e44b2227103c9ef21dbdb 
+contract KreditsKit is KitBase  {
 
+    // ensure alphabetic order
+    enum Apps { Contribution, Contributor, Proposal, Token }
```
fsmanuel: ❤️

scripts/deploy-kit.js
fsmanuel: Much better with a normal function!

scripts/new-dao.js
fsmanuel: I was thinking about `updateAddresses` but as it also updates IDs I think `fileInject` works fine.

scripts/deploy-kit.js
```diff
@@ -0,0 +1,68 @@
...
+  const networkPromise = new Promise((resolve, reject) => {
+    web3.version.getNetwork((err, network) => {
+      if (err) {
+        reject(err);
+      } else {
+        resolve(network);
+      }
+    })
+  })
+  const networkId = await networkPromise;
```
fsmanuel: const networkId = await getNetworkId(web3);

scripts/new-dao.js
```diff
@@ -0,0 +1,53 @@
...
+  const networkPromise = new Promise((resolve, reject) => {
+    web3.version.getNetwork((err, network) => {
+      if (err) {
+        reject(err);
+      } else {
+        resolve(network);
+      }
+    })
+  })
+  const networkId = await networkPromise;
```
fsmanuel: const networkId = await getNetworkId(web3);

scripts/deploy-kit.js
skddc: Have you tried the new suggestion feature in these comments? It's pretty useful!

scripts/seeds.js
```diff
@@ -6,46 +6,41 @@ const Kredits = require('../lib/kredits');
...
-
-  Registry.deployed().then(async (registry) => {
-
-    const networkId = parseInt(web3.version.network);
-    const provider = new ethers.providers.Web3Provider(
-      web3.currentProvider, { chainId: networkId }
-    );
-    const kredits = await Kredits.setup(provider, provider.getSigner());
+module.exports = async function(callback) {
+  const networkId = parseInt(web3.version.network);
```
fsmanuel: const networkId = await getNetworkId(web3);

scripts/current-address.js
```diff
@@ -0,0 +1,22 @@
+const knownAddresses = require('../lib/addresses/dao.json');
+
+module.exports = function(callback) {
+
+  console.log('# All known addresses');
+  Object.keys(knownAddresses).forEach((networkId) => {
+    console.log(`  Network ID: ${networkId} => ${knownAddresses[networkId]}`);
+  })
+
+  const networkId = web3.version.network;
```
fsmanuel: `const networkId = await getNetworkId(web3);`

lib/kredits.js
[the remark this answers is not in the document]
fsmanuel: 👍

scripts/helpers/networkid.js
```diff
@@ -0,0 +1,11 @@
+module.exports = function(web3) {
+  return new Promise((resolve, reject) => {
+    web3.version.getNetwork((err, network) => {
+      if (err) {
+        reject(err);
+      } else {
+        resolve(network);
```
fsmanuel: Do the address files need the `networkId` as a string? Otherwise I would do the `parseInt` here.

bumi: yeah... that keeps bothering me also. but I don't know... maybe we sometimes use strings maybe integers. 
I think I should go through the whole code once and make sure we use strings everywhere. 
but also I am nut sure what web3 or ethers.js return as network id - if that is consistent.

fsmanuel:
@bumi luckily we didn't implement the diff ratio factor. You would be doomed: +70,784 −2,006

bumi:
@fsmanuel yeah, I hoped nobody (and especially you) doesn't notice it :D

bumi:
and we still should implement that... hopefully after this one is merged.. lol :D

lib/addresses/dao.json
```diff
@@ -0,0 +1 @@
+{"4":"0x95a7ce185efc2d1f13efd2a00ee9247c51ea7009","14945560":"0x183af3950364390a266edff2a0e7c4c2f95c0691","23827572":"0xe4e0e7fe54d9189df29a80c07ab733fc9a212761","42097210":"0x183af3950364390a266edff2a0e7c4c2f95c0691","51657314":"0xad3a80686847d55d7b14b930bc6db53681ae1b1e","55632786":"0x4fde16c57ddf6d4870d5edd599074bb50dc96f88","65047207":"0xb67567175ac213f6f622b23d3d972d5b877c4813","74971415":"0x183af3950364390a266edff2a0e7c4c2f95c0691","85406806":"0x192db9a1f757c7c2afd03cb5afe883b27d0d3937"}
```
bumi: I am wondering how I can prevent that those local address details get commited. 
Those "14945560" network IDs are devchain IDs. 
so far I don't think it is worth to write code for that and maybe simply try to ignore commiting those changes, but if somebody has an idea...?

bumi:
ok, now we should soon merge it because I am adding stuff that really should not be here :/ :D

lib/addresses/dao.json
fsmanuel: Maybe a git pre commit hook would work. I have never worked with git hooks...

skddc: Yes, I think that could be a good solution. Git hooks are pretty easy to work with actually.

bumi: hmm. the issue is locally you need those addresses in the file but in the repo we only need the commonly used ones. 
so what would the hook do? 

then it probably does not change that often...

bumi:
now [frame](https://frame.sh/) and potentially also local nodes are supported for deployment and for the helper scripts .
using infura the helper scripts now also have a readonly mode to testnet/mainnet deploys. 
e.g. `truffle exec scripts/list-proposals.js --network=rinkeby`  will list all proposals from the rinkeby testnet.

scripts/add-contribution.js
```diff
@@ -1,16 +1,15 @@
 const promptly = require('promptly');
 
-const ethers = require('ethers');
-const Kredits = require('../lib/kredits');
-
-const getNetworkId = require('./helpers/networkid.js')
+const initKredits = require('./helpers/init_kredits.js');
```
fsmanuel: ❤️

lib/kredits.js
```diff
@@ -81,6 +81,10 @@ class Kredits {
     return this.contractFor('Proposal');
   }
 
+  get Operator() {
+    return this.Proposal();
```
fsmanuel: You don't need the `()` see `get Contributors()`

bumi: ah ja... thx.

scripts/add-contribution.js
bumi: is it actually `init_kredits.js` or initKredits.js` as file name?

bumi:
I've updated ethers.js here now... as ethers.js is also a kredits-web dependency we have to update it there. 
I should really wait until we merge this PR....

scripts/add-contribution.js
fsmanuel: I think JS folks use `require('./helpers/init-kredits');` You don't need the `.js`

fsmanuel:
Ship it!

skddc:
Went through the instructions and bootstrap etc. worked like a charm!

bumi:
@skddc if the last changes solve the problem of the changing  network id then merge it. \o/

skddc:
bluesbrothers

PR #63, opened 2019-04-03
Fix contributor oracle permission auth

bumi:
It seems that the entity is either defined by the permission or if we want to use the oracle the permission must be defined for any_entity.
In that case the oracle does not get the msg.sender as who/entity thus we will use tx.origin in that case.


skddc:
BAAM! :+1:

PR #67, opened 2019-04-03
Add balance when loading contributor

bumi:
This is a regression introduced by the latest contract updates.
Now we return the balance again when loading contributor data.


skddc:
Which balance though? Kredits that have been withdrawn, or kredits that have been earned? (I'd prefer the latter.)

bumi:
this PR only brings back the functionality that we had before. It returns the ERC20 kredits balance of the user. 

Though I am wondering if we should rather load the balances (kredits and contribution) in the client/kredits-web. Would keep the contract code also smaller and less coupled.

skddc:
> Though I am wondering if we should rather load the balances (kredits and contribution) in the client/kredits-web. Would keep the contract code also smaller and less coupled.

But it also increases the amount of HTTP requests for every loading of the app. Not sure which one is better.

bumi:
yep, that's why I made this PR.

skddc:
It werks! :tada:

PR #69, opened 2019-04-03
Fix permissions to manage contributors

bumi:
Uses the contributor oracle to decide if somebody has the MANAGE_CONTRIBUTOR_ROLE

note: 

this changes the KreditsKit, which means only new DAOs will have the new permissions set by default. 
there is no nice script yet to manage the advanced roles with permissions. (this should probably be somehow added to the aragon cli) 

So either you somehow do that manually (which I have never done so far) or you just create a new DAO (in development)


skddc:
:+1:

PR #73, opened 2019-04-04
Convert contribution owner to ID, use smaller number formats

skddc:
Sorry for mixing these two topics into one PR. This changes the contribution to be tied to a contributor ID instead of an account address. It also changes a bunch of numbers from uint256 to uint32.


apps/contribution/contracts/Contribution.sol
```diff
@@ -4,21 +4,29 @@ import "@aragon/os/contracts/apps/AragonApp.sol";
...
+
   // ensure alphabetic order
   enum Apps { Contribution, Contributor, Proposal, Token }
   bytes32[4] public appIds;
 
   struct ContributionData {
-    address contributor;
-    uint256 amount;
+    uint32 contributorId;
+    uint32 amount;
```
fsmanuel: The `mintFor` takes a `uint256`

lib/contracts/contribution.js
```diff
@@ -20,24 +19,40 @@ class Contribution extends Base {
...
-      .then((data) => {
+      .then(data => {
         return this.ipfs.catAndMerge(data, ContributionSerializer.deserialize);
       });
 
   }
 
-  getByContributor(contributor) {
-    return this.functions.balanceOf(contributor)
+  getByContributorId(contributorId) {
```
fsmanuel:
```suggestion
getByContributorId(contributorId) {
  return this.functions.getContributorAddressById(contributorId)
    .then(address => this.getByContributorAddress(address));
}
```

lib/contracts/contribution.js
```diff
@@ -20,24 +19,40 @@ class Contribution extends Base {
...
+    //         });
+    //     }
+    //
+    //     return RSVP.all(contributions);
+    //   });
+  }
+
+  getByContributorAddress(address) {
+    return this.functions.balanceOf(address)
       then((balance) => {
```
fsmanuel: It is missing a `.`

apps/contribution/contracts/Contribution.sol
skddc: Yes, but we're converting it before minting in `claim()`. Nobody has to mine more than 2.2 billion kredits at once, so the locally stored amount can be a smaller format.

lib/contracts/contribution.js
skddc: I may be blind, but where exactly? Maybe you could turn it into a suggestion.

lib/contracts/contribution.js
```diff
@@ -1,20 +1,20 @@
...
 class Contribution extends Base {
   all() {
     return this.functions.contributionsCount()
-      .then(count => {
+      .then(async (count) => {
         let contributions = [];
 
         for (let id = 1; id <= count; id++) {
-          contributions.push(this.getById(id));
+          const contribution = await this.getById(id)
```
fsmanuel: ❤️

bumi: to promise or not to promise? :bumiwonders:

fsmanuel: If you can wait for it you don't need to promise anything.

skddc:
fuck-yeah

PR #76, opened 2019-04-05
Dynamic function to test for core contributor flag

bumi:
This removes the isCore flag and allows us to dynamically calculate the
core flag for contributors.

For now this is just us (the first 6 contirbutors)

Also we do not need the default contributor anymore because the deploy
user has the role to manage contributors and can create the first real
contributors.


fsmanuel:
As far as I can tell that looks good!

skddc:
thumbs-up-machine

PR #78, opened 2019-04-05
Set contribution veto period to 1 week

bumi:
This makes a contribution claimable after roughly 7 days. before that core contributors can veto the contribution. 

update: 
the first 9 contributions are instantly claimable!!! this is helpful in development... 
not sure if we should keep that. ideally we will make the period configurable.


skddc:
thumbs-up-asian-guy-gif

bumi:
yeah, like...

PR #80, opened 2019-04-06
Rename ipfsHash to hashDigest in Contributor

### LINKED ISSUE #75: Unify hashDigest
Can we address this issue as well:

https://github.com/67P/kredits-contracts/blob/93e0b30b357e69a892b0c3187a813b95f0a1d619/lib/contracts/contributor.js#L24-L25

I have ~17~ 9 `ipfsHash` matches in https://github.com/67P/kredits-contracts/blob/master/apps/contributor/contracts/Contributor.sol

Update:
I think we don't use this function at all:
https://github.com/67P/kredits-contracts/blob/93e0b30b357e69a892b0c3187a813b95f0a1d619/apps/contributor/contracts/Contributor.sol#L72-L80

bumi:
naming should be consistent with the other contracts and digest is more correct.

rebased version of #79 
closes #75 - do you see anything missing @fsmanuel?


apps/contributor/contracts/Contributor.sol
```diff
@@ -60,23 +60,23 @@ contract Contributor is AragonApp {
     ContributorAccountUpdated(id, oldAccount, newAccount);
   }
 
-  function updateContributorIpfsHash(uint32 id, bytes32 ipfsHash, uint8 hashFunction, uint8 hashSize) public isInitialized auth(MANAGE_CONTRIBUTORS_ROLE) {
+  function updateContributorIpfsHash(uint32 id, bytes32 hashDigest, uint8 hashFunction, uint8 hashSize) public isInitialized auth(MANAGE_CONTRIBUTORS_ROLE) {
```
fsmanuel: Why do we need that? If we need it I would rename it to `updateContributorProfile` like the event.

bumi: very true. 
I think we need it to be able to update the user's details - like profile picture or whatever.

skddc: Any and all user details actually. Everything except the Ethereum account/address.

apps/contributor/contracts/Contributor.sol
```diff
@@ -60,23 +60,23 @@ contract Contributor is AragonApp {
...
     Contributor storage c = contributors[id];
-    bytes32 oldIpfsHash = c.ipfsHash;
-    c.ipfsHash = ipfsHash;
+    bytes32 oldHashDigest = c.hashDigest;
+    c.hashDigest = hashDigest;
     c.hashFunction = hashFunction;
     c.hashSize = hashSize;
 
-    ContributorProfileUpdated(id, oldIpfsHash, c.ipfsHash);
+    ContributorProfileUpdated(id, oldHashDigest, c.hashDigest);
```
fsmanuel: General question: When do you need to use `emit` for an event?

bumi: always in newer solidity version. events without emit is deprecated - so we must update that soon.

fsmanuel: Thanks!

fsmanuel:
👍

PR #84, opened 2019-04-08
WIP: Update all package-lock.json files

fsmanuel:


bumi:
:+1:  I don't understand why those lock files always change. seems to me somehow that everyone of use gets a different one after installation.

PR #86, opened 2019-04-09
Don't run `npm install` on bootstrap

skddc:
When one wants to install, they can install. When one wants to bootstrap, they should be able to only bootstrap.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "repl": "truffle exec scripts/repl.js",
     "seeds": "truffle exec scripts/seeds.js",
     "compile-contracts": "aragon contracts compile --all",
-    "bootstrap": "./scripts/every-app.sh \"npm install \" && npm run reset:hard && npm run seeds",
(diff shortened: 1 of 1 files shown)


bumi:
we should probably remove the `postinstall` script also? 
@fsmanuel as you have recently done the setup with out much additional explanation, what do you think is the best and easiest setup flow?

skddc:
Can we merge this first? I have another PR coming for postinstall.

bumi:
my thoughts when making `bootstrap` were that this runs once to set everything up. 
after that one can use `reset:*` to get fresh DAOs and seeds to fill it optionally with basic data. 

The alternative to manually cd into every directory and install the dependencies would then be to remove the package.json and the need to run `npm i` in every app?

skddc:
> The alternative to manually cd into every directory and install the dependencies would then be to remove the package.json and the need to run `npm i` in every app?

No, you can simply use a command for that. It doesn't have to be automatic as postinstall.

PR #87, opened 2019-04-09
Rename postinstall script

skddc:
Prevents it being run when contracts are installed as dependency.


bumi:
:+1: can you update the readme accordingly?

PR #88, opened 2019-04-09
Fix confirmed contribution check

bumi:
If the blocknumber is higher or equal the contribution.confirmedAt block
number then the contribution is confirmed and can be claimed.


skddc:
:+1:

PR #91, opened 2019-04-10
Store kredits tokens as wei with 18 decimals

bumi:
This stores the tokens with 18 decimals as most other tokens and as
assumed by a potential future token standard EIP 777.

This changes the return value of the balance of contributors which would need to be converted to "ether" in the client.

### DIFF
--- a/apps/token/contracts/Token.sol
+++ b/apps/token/contracts/Token.sol
@@ -14,11 +14,15 @@ contract Token is ERC20Token, AragonApp {
 
   function initialize(bytes32[4] _appIds) public onlyInit {
     appIds = _appIds;
+    name = 'Kredits';
+    symbol = 'K';
+    decimals = 18;
     initialized();
   }
 
   function mintFor(address contributorAccount, uint256 amount, uint32 contributionId) public isInitialized auth(MINT_TOKEN_ROLE) {
-    _mint(contributorAccount, amount);
+    uint256 amountInWei = amount.mul(1 ether);
+    _mint(contributorAccount, amountInWei);
     emit LogMint(contributorAccount, amount, contributionId);
   }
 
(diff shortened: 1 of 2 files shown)


apps/token/contracts/Token.sol
```diff
@@ -14,11 +14,15 @@ contract Token is ERC20Token, AragonApp {
 
   function initialize(bytes32[4] _appIds) public onlyInit {
     appIds = _appIds;
+    name = 'Kredits';
+    symbol = 'K';
```
skddc: I think we need more than one letter for a symbol tho.

bumi: what do you suggest? 
so far it was empty and not used.

skddc: I don't think a single person's suggestion is enough to decide this. We need options and ask everyone which one they like best I guess.

bumi: we could add a method in the future to update it. so we can leave that right now and deploy, what do you think? 
in the UI we don't use that entry anyway. it would be used in wallet apps that display token balances, but those entries are also mostly customizable.

skddc: Adding a method vs. just coming up with a decent one now: is there really a difference in time investment?

bumi: it is doing nothing now and later adding a method to change the name if we want, or looking for options and asking everyone now.  :) 

if we like to store the amount as that standard suggests, then we should do that before deploying and not require some data migration later.

skddc: Oh, you mean add a method later. Got it.

Then let's just change it to what we have in the Web UI right now: K$ (as in **k**redit**s**). Or KS without the dollar sign. WDYT?

bumi: :+1: 
that's `₭S` ? 
https://github.com/67P/kredits-web/pull/100/files#diff-816a93e6bfad9d8ac961b7e0174dcf7dR8

skddc: Oh, right. (Although it may actually be different between that list and the toplist, in which case I'd prefer this one.)

Issue #30, opened 2018-04-19
Validate Contributor/Contribution details using json schema


bumi:
The Contributor/Operator class currently does not validate the data against the json schema from kosmos-schemas. To prevent storing invalid data we should validate those before uploading to ipfs and to the smart contract.

skddc:
¡Muy importante!

skddc:
Btw, it's only one function call, documented nicely in the schemas doc: https://github.com/67P/kosmos-schemas#usage

skddc:
Bumping this one. We already have some invalid documents in IPFS docs linked from the Kovan data.

skddc:
All done.

PR #92, opened 2019-04-10
Validate contribution docs against schema

skddc:
Currently requires [an open PR branch](https://github.com/67P/kosmos-schemas/pull/5) for the schemas, which is adding date and time for contributions.

refs #30

### DIFF
--- a/lib/contracts/contribution.js
+++ b/lib/contracts/contribution.js
@@ -1,5 +1,20 @@
 const ethers = require('ethers');
 
+const schemas = require('kosmos-schemas');
+const tv4 = require('tv4');
+const validator = tv4.freshApi();
+
+validator.addFormat({
+  'date': function(value) {
+    const dateRegexp = /^[0-9]{4,}-[0-9]{2}-[0-9]{2}$/;
+    return dateRegexp.test(value) ? null : "A valid ISO 8601 full-date string is expected";
+  },
+  'time': function(value) {
+    const timeRegexp = /^([01][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9]|60)(\.[0-9]+)?(([Zz])|([\+|\-]([01][0-9]|2[0-3]):[0-5][0-9]))$/;
+    return timeRegexp.test(value) ? null : "A valid ISO 8601 full-time string is expected";
+  }
+})
+
 const ContributionSerializer = require('../serializers/contribution');
 const Base = require('./base');
 
@@ -48,12 +63,15 @@ class Contribution extends Base {
   }
 
   addContribution(contributionAttr, callOptions = {}) {
-    let json = ContributionSerializer.serialize(contributionAttr);
-    // TODO: validate against schema
+    let jsonStr = ContributionSerializer.serialize(contributionAttr);
+
+    // Validate JSON document against schema
+    let result = validator.validate(JSON.parse(jsonStr), schemas['contribution']);
+    if (!result) { return Promise.reject(validator.error); }
 
     return this.ipfs
-      .add(json)
-      .then((ipfsHashAttr) => {
+      .add(jsonStr)
+      .then(ipfsHashAttr => {
         let contribution = [
           contributionAttr.amount,
           contributionAttr.contributorId,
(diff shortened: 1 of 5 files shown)


lib/contracts/contribution.js
```diff
@@ -1,5 +1,20 @@
...
+validator.addFormat({
+  'date': function(value) {
+    const dateRegexp = /^[0-9]{4,}-[0-9]{2}-[0-9]{2}$/;
+    return dateRegexp.test(value) ? null : "A valid ISO 8601 full-date string is expected";
+  },
+  'time': function(value) {
+    const timeRegexp = /^([01][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9]|60)(\.[0-9]+)?(([Zz])|([\+|\-]([01][0-9]|2[0-3]):[0-5][0-9]))$/;
+    return timeRegexp.test(value) ? null : "A valid ISO 8601 full-time string is expected";
+  }
+})
```
bumi: afaik we use the same format for the proposals and a proposal once accepted creates a contribution. 
So we should probably add this also in the proposal wrapper.

scripts/add-contribution.js
```diff
@@ -26,23 +27,32 @@ module.exports = async function(callback) {
...
-  });
+  kredits.Contribution.addContribution(contributionAttributes, { gasLimit: 300000 })
+    .then(result => {
+      console.log("\n\nResult:");
+      console.log(result);
+      callback();
+    })
+    .catch(error => {
+      console.log('Failed to create contribution');
+      callback(inspect(error));
```
bumi: does this work? thought the callback should get an error object.

lib/contracts/contribution.js
skddc: Yeah, but proposals aren't really part of MVP.

scripts/add-contribution.js
skddc: We need to reject with the tv4 error object either way. It does show the error and doesn't stall. So it works well enough to be usable, yes.

lib/contracts/contribution.js
bumi: yeah, though it might be worth to move this into a separate file that exports validator 
and then do a `const validator = require('validator')` ?

lib/contracts/contribution.js
```diff
@@ -48,12 +63,15 @@ class Contribution extends Base {
   }
 
   addContribution(contributionAttr, callOptions = {}) {
-    let json = ContributionSerializer.serialize(contributionAttr);
-    // TODO: validate against schema
+    let jsonStr = ContributionSerializer.serialize(contributionAttr);
+
+    // Validate JSON document against schema
+    let result = validator.validate(JSON.parse(jsonStr), schemas['contribution']);
```
bumi: I somewhat dislike that we do a `JSON.stringify` in the ContributionSerializer and then do a `JSON.parse` again to validate it. 
maybe we can have two function in the ContributionSerializer? 
of even do the validation in the ContributionSerializer? 

something like: 
```js
const c = Contribtion.new(contribtionAttr);
c.isValid() // calls the validator.validate() function
c.error // returns validator.error
c.serialize() // same as now
```

but we can refactor that later I guess.

skddc: Yes, makes sense. We can refactor it when adding validation for contributor details.

fsmanuel: Maybe something like:

```js
// This tv4 api feels strange...
const [valid, validator] = ContributionSerializer.validate(contribtionAttr);
if (!valid) { return Promise.reject(validator.error); }
```
`.validate` should use `_serialize()` to do the work and `serialize()` can use the private method as well but returns the JSON string...

I think the overhead of `_serialize()` is not much so we can do it twice.

lib/contracts/contribution.js
fsmanuel: I think it makes a lot of sense to have everything in the serializers. Still not sure how best to implement the validation in the serializer...

lib/contracts/contribution.js
skddc: Already came up with an even cleaner solution. Promisify ALL THE THINGS. ;)

skddc: https://github.com/67P/kredits-contracts/blob/d953141f52d515dbcba4216c5eab9111c58467a2/lib/contracts/contribution.js#L51-L54

PR #93, opened 2019-04-10
Production release to rinkeby

### LINKED ISSUE #90: Deploy production org/contracts

bumi:
Deployed all apps and created a new Kredits DAO. 

here is our address: 0xc34edf7d11b7f8433d597f0bb0697acdff55ef14

ToDo: 

- [x] add contributors

closes #90


bumi:
yay! 
once we have hubot deployed we need to set the permissions for hubot's wallet. (guess we should create a new one three?)

PR #94, opened 2019-04-11
Refactor contribution serializer and validation

skddc:
And fix seeds.

### DIFF
--- a/lib/serializers/contribution.js
+++ b/lib/serializers/contribution.js
@@ -1,45 +1,24 @@
+const schemas = require('kosmos-schemas');
+const validator = require('../utils/validator');
+
 /**
- * Handle serialization for JSON-LD object of the contribution, according to
- * https://github.com/67P/kosmos-schemas/blob/master/schemas/contribution.json
+ * Serialization and validation for JSON-LD document of the contribution.
  *
  * @class
  * @public
  */
 class Contribution {
- /**
-  * Deserialize JSON to object
-  *
-  * @method
-  * @public
-  */
-  static deserialize(serialized) {
-    let {
-      date,
-      time,
-      kind,
-      description,
-      details,
-      url,
-    } = JSON.parse(serialized.toString('utf8'));
 
-    return {
-      date,
-      time,
-      kind,
-      description,
-      details,
-      url,
-      ipfsData: serialized,
-    };
+  constructor(attrs) {
+    Object.keys(attrs).forEach(a => this[a] = attrs[a]);
   }
 
- /**
-  * Serialize object to JSON
-  *
-  * @method
-  * @public
-  */
-  static serialize(deserialized) {
+  /**
+   * Serialize object to JSON
+   *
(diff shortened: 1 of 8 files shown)


skddc:
I set this to WIP, because I'm going to add proposal validation and valid proposal seeds to this as well.

skddc:
Ready for review!

lib/serializers/contribution.js
```diff
@@ -1,45 +1,24 @@
...
-  * @method
-  * @public
-  */
-  static serialize(deserialized) {
+  /**
+   * Serialize object to JSON
+   *
+   * @public
+   */
+  serialize () {
```
fsmanuel: What about:

```js
  serialize (stringify=true) {

    if (stringify) {
      return JSON.stringify(data, null, 2);
    } else {
      return data;
    }
  }

  validate () {
    const serialized = this.serialize(false);
    const valid = validator.validate(serialized, schemas['contribution']);
    return valid ? Promise.resolve() : Promise.reject(validator.error);
  }
```

skddc: Let's try ro create smaller PRs for suggestions like this, instead of having them hold up larger branches from being merged.

(E.g. personally, I don't like unnamed options, so I'd probably propose to use an options object for this. But that's a completely different discussion than if the validation here works or not.)

What do you think?

fsmanuel: Well I think that the initial idea to move the validation into the serializer was to get rid of the `JSON.stringify` ->`JSON.parse` thing for the validation. My suggestion is focusing on how we can prevent this.
Another thing I don't understand is why you removed the static functions and don't use the class at all.

skddc: > Well I think that the initial idea to move the validation into the serializer was to get rid of the `JSON.stringify` ->`JSON.parse` thing for the validation. My suggestion is focusing on how we can prevent this.

That's not at all how I understood the idea. I merely refactored everything to be both shared modules and also do the validation in the class instead of outside.



> Another thing I don't understand is why you removed the static function and don't use the class at all.

No idea what that means. It's using only the class now, because I added more functions to it. I turned the static function into a prototype function, because that's actually using the class with prototype properties then.

skddc: By the way, in a way I also like validating *exactly* what's being stored in IPFS. I don't see it as a problem, and I have no idea why it's keeping us from merging now.

fsmanuel: I was referring to this https://github.com/67P/kredits-contracts/pull/92#discussion_r274168386
So my suggestion adds little to archive this. I modified to code above to better illustrate the idea. If you disagree merge it as is.

> Let's try to create smaller PRs for suggestions like this, instead of having them hold up larger branches from being merged.

I think reviews are the place to raise this discussions, aren't they?

fsmanuel: That being said. I like your implementation!

PR #96, opened 2019-04-12
Add pagination for .all methods

fsmanuel:
I removed rsvp as a dependency as we only use Promise.all

Since the setup only works for me with yarn it would be great if someone can run `npm install` and push the package-lock.json into this branch.


lib/utils/pagination.js
```diff
@@ -0,0 +1,46 @@
...
+  let end = offset + size;
+  if (end > recordCount) {
+    let diff = end - recordCount;
+    size = size - diff;
+  }
+
+  return Array.from({ length: size }, mapFunction);
+}
+
+module.exports = function paged(recordCount, options = {}) {
```
bumi: the call would be somthing like: 
`page(100, { order: 'desc', page: { size: 10, number: 1} })` ?

was wondering if that is less verbose? 
`page(100, { order: 'desc', per_page: 10, page: 1 })`  - but that's probably my ruby will_paginate head :D

fsmanuel: Yeah, I'm not sure which one is better. I took it from [JSON:API spec](https://jsonapi.org/format/#fetching-pagination) but I don't care how we do it.
@skddc what is your preference?

fsmanuel:
One thing I noticed is that we don't return any information on how many records/pages we have. This is not a problem if we use infinite scrolling as the code always checks that the results are in range. But if we want to have a pagination with pages in the frontend that might be a problem. We can always use `functions.*Count` but would duplicate logic...
To make it easier I'll add a `count` property on the contract and we can iterate from there.

fsmanuel:
Updated the code. Not sure if `Record` is the best name...

skddc:
Very cool stuff. I'll work on using the contracts in production before getting to this, so that the kredits on these PRs can actually be recorded. ;)

lib/utils/pagination.js
bumi: ah ok, then I guess JSON:API spec is better!

bumi:
I think we can optimize if we run in that problem, so a `count` function should be enough now. 

so let's make this the first recorded github contribution on or rinkeby production org. 
I will merge this once hubot is deployed and configured...(and I hope it then works :D)

fsmanuel:
@bumi What do you need for my contributor profile? Can we take my old one?

bumi:
see contributor schema.
ethereum address, name, github user+id, maybe wiki username

PR #98, opened 2019-04-12
WIP: Use yarn instead of npm

fsmanuel:
Using yarn fixed my setup problems.
In the beginning I had a problem with a web3 package but this error is somehow gone, now.
Can someone test if it works for you as well?


fsmanuel:
It has some historical reasons why yarn exists. npm2 installed all the packages nested aka the same package got installed multiple times. Yarn on the other hand flattend the dependencies and did a much better job at version resolution. It also installed the packages in parallel and was much faster. Since then npm3 also flattend the dependencies but is still very slow. The yarn lock file is much more stable and doesn't change every time. So I prefer yarn over npm for package management...

bumi:
ok, I see.. but `npm` works as it does currently.. so it is not a breaking change, is it?

skddc:
Except for the difference in lockfiles, npm over time introduced pretty much any and all benefits that yarn ever had over npm afaik. I think our code should definitely support being used with yarn, but I'd propose to keep npm as the default, simply because it is the de-facto standard package manager for JavaScript, and yarn is just an alternative client for npm's registry.

fsmanuel:
Alright! I'll rebase and readd the package-lock files so we can use both.

fsmanuel:
Superseded by #102

PR #104, opened 2019-04-16
Update kosmos-schema dependency

bumi:
kosmos-schema v2.0.0 is released


bumi:
I merge this as it is currently broken.

PR #106, opened 2019-04-17
[WIP] Solidity contracts tests

haythem96:
Unit tests for Kredits smart contracts


fsmanuel:
@haythem96 pretty cool to see how to test the contracts. I'm wondering why the `pull upstream master` commit reverts the pagination changes? I guess you need to rebase/merge master again...

haythem96:
for what [this](https://github.com/haythem96/kredits-contracts/blob/5368a3d7d819bf498ca7b0f1d9a690ffacf1a48c/apps/contributor/contracts/Contributor.sol#L34) unused variable ?

bumi:
@haythem96 you mean `root`? I missed that in this commit: https://github.com/haythem96/kredits-contracts/commit/766463b57b87b2233a5bd681310da7332b9db904#diff-b24176ed2ec209da948c7ef4f480abe1L36
but the kit is still calling the function like that.

apps/contributor/test/contributor.js
```diff
@@ -0,0 +1,160 @@
...
+            return assertRevert(async() => {
+              await contributor.updateContributorAccount(id.toNumber(), oldAccount, ZERO_ADDR);
+              'contributor does not exist'
+            });
+        });
+
+        it('update contributor profile hash', async() => {
+            await contributor.updateContributorProfileHash(id.toNumber(), hashDigest, hashFunction, hashSize);
+            let contributorId = await contributor.getContributorIdByAddress(oldAccount);
+            assert.equal(contributorId.toNumber(), 0);
```
skddc: This doesn't actually check for an updated hash afaics.

haythem96: yup...solved in this [commit](https://github.com/67P/kredits-contracts/pull/106/commits/ab6d6a40251a9589d99886b0e53bb86b74f50080)

apps/contribution/test/contribution.js
```diff
@@ -0,0 +1,196 @@
...
+        //create token mint permission for coin owner
+        await acl.createPermission(
+            root,
+            token.address,
+            await token.MINT_TOKEN_ROLE(),
+            root,
+            { from: root }
+        );
+
+        //acl.grantPermission(contribution, token, await token.MINT_TOKEN_ROLE(), {from: root});
```
haythem96: I have a problem here granting the mint token permission to the contribution app...

apps/proposal/test/proposal.js
```diff
@@ -0,0 +1,156 @@
+const namehash = require('eth-ens-namehash').hash;
+
+const Proposal = artifacts.require("Proposal.sol");
+const { Contributor, getContributorContract } = require("../../contributor/artifacts");
```
haythem96: idk why I have an error here importing other apps contracts?

apps/contribution/test/contribution.js
bumi: what do you mean? what's the error?

apps/proposal/test/proposal.js
bumi: looks to me like `artifacts` might not defined in the contributor/artifacts.js file. 
`artifacts.require('Contributor')` does not work here? because the proposal app does not know about the Contributor.sol?

from where is truffle loading the contracts?

apps/contribution/test/contribution.js
haythem96: a revert...

apps/proposal/test/proposal.js
haythem96: I did the same with the Contribution app to access to the Token app contracts, and it worked...
If I understand correctly `artifacts.require('Contributor.sol')` in Contributor app should b able to access that contract inside that app contracts folder... or no ?

fsmanuel:
@haythem96 You should run `npm run lint:wrapper --fix` and `npm run lint:contract-tests --fix` to get the fixable lint errors fixed. After that see what lint errors are left and fix them by hand.

haythem96:
> `npm run lint:contract-tests --fix`

I did that...had some issue with the lint rules(e.g in the contracts test it show `'artifacts' is not defined` or `'before' is not defined`... so I just disabled some rules in the tests files

haythem96:
Contributor app tests works... still have issue with Contribution app, the `getTokenContract()` always return `address(0)` in the tests

skddc:
This is turning into quite the monolithic monster pull request imo.

@haythem96 You think we can split this up, so it's easier to review, merge, and collaborate?

For example I could imagine to first just merge the basic test suite setup, so that the test command simply works and it's possible to add tests. Then we could merge whatever tests you already have (I guess Contributor), and then steadily add more to cover all code.

What do you think?

skddc:
P.S.: The way kredits are earned with the labels, it should actually encourage the smallest possible pull requests, because the highest kredits amount is "large" (`kredits-3`), which shouldn't be more than a few days of work (equivalent to 10 small contributions, or a bit more than 3 medium ones). And in this case, I'm certain it should be more than a single large one.

haythem96:
okay will close this one currently & do others small PRs

skddc:
Cool!

PR #108, opened 2019-04-19
Add kredits-formatter and additionally format balance

bumi:
This adds a `balanceInt` value to the contributor data that has the
token balance formatted as full Kredits. (similar to Ether)

any naming suggestions?


lib/contracts/contributor.js
```diff
@@ -1,5 +1,6 @@
 const Record = require('./record');
 const ContributorSerializer = require('../serializers/contributor');
+const KreditsFormatter = require('../utils/kredits-formatter');
```
skddc: As this is not a class, but just one formatting function, I'd calls the file `format_kredits` and import it as `formatKredits()`.

lib/contracts/contributor.js
```diff
@@ -8,9 +9,13 @@ class Contributor extends Record {
 
   getById(id) {
     return this.functions.getContributorById(id)
-      .then((data) => {
+      .then(data => {
         return this.ipfs.catAndMerge(data, ContributorSerializer.deserialize);
-      });
+      })
+      .then(data => {
+        data.balanceInt = KreditsFormatter(data.balance);
```
skddc: What's the benefit of doing this here instead of before the return in the first `this` above?

lib/contracts/contributor.js
```diff
@@ -8,9 +9,13 @@ class Contributor extends Record {
...
+        return Promise.resolve(data);
```
fsmanuel: No need for `Promise.resolve` because every `then` returns a new Promise.

`return data;` should work.

lib/contracts/contributor.js
fsmanuel: I think the convention is a dash for filenames in JS.
👍 `formatKredits()`

lib/contracts/contributor.js
```diff
@@ -8,9 +9,13 @@ class Contributor extends Record {
...
+      })
```
fsmanuel: `;`

lib/contracts/contributor.js
bumi: `catAndMerge` returns a promise, so I only can do it in a next `then`, no?

skddc: I thought `getContributorById` already includes the balance. In which case it doesn't matter what's returned afterwards, as it would already include the formatted balance beforehand.

bumi: ah, true. before passing it to IPFS. :+1:

skddc:
There's a merge conflict here.

bumi:
conflict resolved.

PR #109, opened 2019-04-19
Handle errors in contributor list script

skddc:


bumi:
fyi. errors might happen also in the `toString()` function. 
the general problem needs to be globally addressed - maybe with a truffle update or with removing truffle's web3 in the scripts and only using the wrapper.

bumi:
I think I merged this too fast and without actually running the code. 
you set the `const contributors` variable inside a try block, so the variable is not available further down? or am I missing something?

skddc:
That could be the actual reason why I wasn't able to list contributors anymore. Sorry! :flushed:

PR #110, opened 2019-04-19
Add contract calls for total contribution balances

bumi:
Those methods return the total kredits amounts of contributions.


apps/contribution/contracts/Contribution.sol
```diff
@@ -118,6 +118,26 @@ contract Contribution is AragonApp {
   // Custom functions
   //
 
+  function totalCount(bool confirmedOnly) public view returns (uint256 count) {
```
skddc: I think this is mixing up "count" and "balance". If it's returning the number of contributions, then it should be "count", but if it's returning the amount of kredits that have been created, then it should be something different. "Balance" also doesn't apply in this case, because that word means an amount per person/account.

Thus, I'd propose to call it something like `totalKreditsAmount` or `totalKreditsSupply` or simply `totalKredits`.

bumi: yeah, naming sucks.. especially because the standard method names are misleading. 

`totalKredits` does not work imo because we might sell kredits tokens - thus the tokens could be more than the ones eared. 

from your ideas I think `totalKreditsEarned` works best I think?

bumi: that then would be `totalKreditsEarned` and `contributorTotalKreditsEarned` ?

skddc: What's `contributorTotalKreditsEarned`? Maybe you mean `Contributor.totalKreditsEarned`?

bumi: if we were not in a smart contract world then maybe. 
but that is a view method on a contract.

we can define a `totalKreditsEarned` on the Contributor wrapper class.

skddc: Sorry, but I still don't understand what it returns in `Contribution` then.

bumi: the total numbers of kredits for that contributor. 
it counts the kredits of all the contributor's contributions.

skddc: OK, thanks. I think the helper function should be the contributor contract, not the wrapper, but it would at least be necessary as a function in the contribution contract still. Reason being that the contributor contract can then deliver toplists sorted by kredits earned and such. Same for list of core team members etc., saving the client from having to load the entirety of the data every time.

skddc:
Missing kredits label btw.

apps/contribution/contracts/Contribution.sol
bumi: try to implement it, you will see that this not possible as you imagine. 
also it would again increase the coupling of the two different contracts. 

In a normal object oriented world I'd agree, but not in that contract world which also makes it hard to read and work with data. (I also think the smart contracts are not made for that and we would need to do such things rather on a different level).

skddc: If it's easy to do like this in the contribution contract, then I don't see why the contributor contract cannot use those functions, as it does for other functions as well. I'm not talking about object orientation at all. Not sure where that's coming from.

bumi: PRs welcome :)

PR #113, opened 2019-04-24
Add Gitea site/username to serializer and seeds

skddc:
Allows to add Gitea account information to contributor profiles.


skddc:
> but I was wondering if we still need to have the account details on the root level of the object and if we can somehow have this universally to be able to add integrations without changing this code here?

Yes, I would like this refactored soon. It should be possible to add arbitrary accounts.

PR #114, opened 2019-04-24
Improve/fix scripts

skddc:
* Improve error handling for list-contributors script
* Fix function name for total kredits issued/earned

### DIFF
--- a/scripts/list-contributors.js
+++ b/scripts/list-contributors.js
@@ -34,13 +34,12 @@ module.exports = async function(callback) {
         c.ipfsHash
       ])
     })
+
+    console.log(table.toString())
   } catch(e) {
     callback(e);
     return;
   }
 
-  console.log(table.toString())
-
   callback()
 }
-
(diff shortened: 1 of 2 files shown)


scripts/list-contributors.js
```diff
@@ -21,24 +21,24 @@ module.exports = async function(callback) {
...
-      `${c.name}`,
-      c.isCore,
-      ethers.utils.formatEther(c.balance),
-      c.totalKreditsEarned.toString(),
-      c.contributionsCount.toString(),
-      c.ipfsHash
-    ])
-  })
-
   console.log(table.toString())
```
bumi: can we add this also in the` try {} catch` block? 
if something fails in this line we would also not see it.

skddc: Feel free to change it.

bumi: fyi: I generally don't like the need for those try/catch blocks everywhere... need to fix the root of the problem and it must be fixed on the truffle level or (as I mentioned what I am working on) that we don't rely on truffle again..

PR #118, opened 2019-04-24
Deprecate old add* methods

fsmanuel:
As we no longer have the operator contract we can deprecate the redundant add methods.
Now it should be clear what is added from the contract name.

I think it's good to deprecate early. The deprecate function could be extended so we can add a version number for the removal.

### DIFF
--- a/lib/contracts/contribution.js
+++ b/lib/contracts/contribution.js
@@ -59,7 +59,7 @@ class Contribution extends Record {
   }
 
   addContribution () {
-    deprecate('The method: `addContribution` is deprecated use `add` instead');
+    deprecate('The function `addContribution()` is deprecated and will be removed in the next major version. Use `add()` instead');
     return this.add(...arguments);
   }
 }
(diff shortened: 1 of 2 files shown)


bumi:
utACK

lib/contracts/contribution.js
```diff
@@ -56,6 +57,11 @@ class Contribution extends Record {
         return this.functions.add(...contribution, callOptions);
       });
   }
+
+  addContribution () {
+    deprecate('The method: `addContribution` is deprecated use `add` instead');
```
skddc:
```suggestion
    deprecate('The function `addContribution()` is deprecated and will be removed in the next major version. Use `add()` instead');
```

lib/contracts/proposal.js
```diff
@@ -35,6 +36,11 @@ class Proposal extends Record {
         return this.functions.addProposal(...proposal, callOptions);
       });
   }
+
+  addProposal () {
+    deprecate('The method: `addProposal` is deprecated use `add` instead');
```
skddc:
```suggestion
    deprecate('The function `addProposal()` is deprecated and will be removed in the next major version. Use `add()` instead');
```

PR #119, opened 2019-04-25
Fix apps lookup

bumi:
We use the directories in ./apps to load available app names.
For this we need to make sure that we only use directories.

Some recent change broke this by adding a file in the apps directory.
So without this fix no DAO can be deployed.

### DIFF
--- a/scripts/deploy-kit.js
+++ b/scripts/deploy-kit.js
@@ -35,7 +35,7 @@ module.exports = async function(callback) {
   }
   console.log(`Using DAOFactory at: ${daoFactory.address}`)
 
-  const apps = fs.readdirSync('./apps')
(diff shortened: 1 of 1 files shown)


scripts/deploy-kit.js
```diff
@@ -35,7 +35,7 @@ module.exports = async function(callback) {
   }
   console.log(`Using DAOFactory at: ${daoFactory.address}`)
 
-  const apps = fs.readdirSync('./apps')
+  const apps = fs.readdirSync('./apps', {withFileTypes: true}).filter(e => e.isDirectory()).map(e => e.name)
```
skddc:
```suggestion
  const apps = fs.readdirSync('./apps', { withFileTypes: true })
                 .filter(e => e.isDirectory()).map(e => e.name);
```

PR #121, opened 2019-04-27
Improve totalKreditsEarned

skddc:
* Use a more reasonable size of integer
* Use better variable names (count != amount, balance != count)


bumi:
fyi: due to the ERC721 standard that naming is a bit confusing e.g. see the `balanceOf` function which does a count of contributions. 
(though I am tending more and more to move away and ignore that standard)

bumi:
Did you test that the different returned types (uint32) don't break the scripts / kredits-web? (I am lazy right now to go through the setup :/ )

skddc:
> Did you test that the different returned types (uint32) don't break the scripts / kredits-web?

No, but it's a simple change for both. If it does break clients, we can release a major version of the module.

skddc:
> fyi: due to the ERC721 standard that naming is a bit confusing e.g. see the `balanceOf` function which does a count of contributions.

Yes, I realize that. But it doesn't mean we have to use words that mean different things for variables in unrelated functions.

bumi:
ToDo: 
- [ ] update ABI

bumi:
we should also update the code to only count non-vetoed contributions and deploy that together on-chain.

skddc:
> we should also update the code to only count non-vetoed contributions and deploy that together on-chain.

Can't see an issue about that. Sounds like a bug to me.

bumi:
update the rest of the code accordingly and added the resulting, updated ABI. 
did not test it with kredits-web though. But I'd say let's merge and deploy this update.

PR #122, opened 2019-04-28
Add script for vetoing via console

skddc:


scripts/veto-contribution.js
```diff
@@ -0,0 +1,32 @@
...
+  try {
+    kredits.Contribution.functions.veto(contributionId, { gasLimit: 300000 })
+      .then(result => {
+        console.log("\n\nResult:");
+        console.log(result);
+        callback();
+      })
+      .catch(error => {
+        console.log('Failed to veto contribution');
+        callback(inspect(error));
```
bumi: why do we have that inspect here - which we do not have in the other scripts? this turns that error into a string.
Because I've never looked into that callback function details? what does it exactly do? is it better to pass in a string than an error object?

skddc: This is copied straight from the other script I added the first error handling to, and where it's still present:

https://github.com/67P/kredits-contracts/blob/master/scripts/add-contribution.js#L56

bumi: ah I see.
I guess we don't need it and also don't use it everywhere.

PR #123, opened 2019-05-03
Fix it

bumi:
seems we did not test it properly and didn't do a good review :)


skddc:
Both the commit title and description as well as the PR title and description don't explain what is being fixed here. Which makes it very hard to review...

bumi:
two missing dots: https://github.com/67P/kredits-contracts/pull/124/files

PR #124, opened 2019-05-03
Fix util import

bumi:


bumi:
merging as it is currently broken. but this should be a `kredits-0` maybe?

skddc:
> but this should be a `kredits-0` maybe?

You did fix it, and you'll have to publish the npm module, too. So I don't see how it's not a valuable small contribution.

bumi:
it was only broken in master, the latest release is still fine as far as I checked.

skddc:
Well, then a core team member you can still merge without the label, if you want.

PR #125, opened 2019-05-03
Fix page number calculation for zero records

bumi:
So far it returned 0 because of those in range checks which both
applied because numberOfPages is 0 if we have no record.

### DIFF
--- a/lib/utils/pagination.js
+++ b/lib/utils/pagination.js
@@ -4,8 +4,8 @@ function pageNumber (number, size, recordCount) {
   number = parseInt(number) || 1;
 
   // Ensure page number is in range
-  number = number < 1 ? 1 : number;
   number = number > numberOfPages ? numberOfPages : number;
+  number = number < 1 ? 1 : number;
 
   return number;
 }


lib/utils/pagination.js
```diff
@@ -4,8 +4,8 @@ function pageNumber (number, size, recordCount) {
   number = parseInt(number) || 1;
 
   // Ensure page number is in range
-  number = number < 1 ? 1 : number;
   number = number > numberOfPages ? numberOfPages : number;
+  number = number < 1 ? 1 : number;
```
skddc: Aside from the fix looking good, are these two lines actually a good idea (as opposed to throwing an exception)?

As far as I understand this code, if I ask for page -2, it will give me page 1 (instead of the second-last e.g.), and if I ask for page 24 out of 23 it will give me 23. But in both cases, I wouldn't know that I got a different page than I asked for, and that the pages I asked for are invalid, right?

bumi: at least in the zero records case I do not want an exception.

skddc: That much is certain. Just asking in general, because I'm not sure it's worth an issue, if I'm not making sense.

bumi: not sure when this would happen.  - we would need then also proper exception handling on the other sides.
need to pick @fsmanuel's thoughts

fsmanuel: It is not optimal but 100% fail save. I guess it depends on how we use it (infinite scroll, normal pagination, etc.). The client somehow needs to implement some of the logic no matter what so I guess the wrapper should just return valid data. I don't like to throw an error here right now because error handling is always a pain.
Ideally we would also return some meta data (current page, page count, per page, etc) so the client can better work with it...

skddc: >  Ideally we would also return some meta data (current page, page count, per page, etc) so the client can better work with it...

:+1:

PR #127, opened 2019-05-14
Allow setting any contributor accounts

bumi:
This allows to pass in an account object when creating or updating a
contribtor.

So far this was only possible by setting the `github_username` etc. attributes.

### DIFF
--- a/lib/serializers/contributor.js
+++ b/lib/serializers/contributor.js
@@ -28,6 +28,7 @@ class Contributor {
       github_username,
       gitea_username,
       wiki_username,
+      accounts
     } = this;
 
     let data = {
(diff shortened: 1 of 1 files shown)


lib/serializers/contributor.js
```diff
@@ -28,14 +28,15 @@ class Contributor {
...
+      accounts
     } = this;
 
     let data = {
       '@context': 'https://schema.kosmos.org',
       '@type': 'Contributor',
       kind,
       name,
-      'accounts': [],
+      'accounts': (accounts || []),
```
skddc:
```suggestion
      accounts: accounts || [],
```

PR #128, opened 2019-05-14
Add helper to initialize a new instance with provider

bumi:
So far we always had to initialze a provider and signer and pass those
to the kredits constructor.
This helper makes it easier to initialize a default ethers provider and
a default signer.

Examples:

```js
Kredits.for({network: 'local'}).init();
Kredits.for({network: 'rinkeby'}, {apm: 'open.aragonpm.eth'}).init();
```

### DIFF
--- a/lib/kredits.js
+++ b/lib/kredits.js
@@ -66,6 +66,23 @@ class Kredits {
     return new Kredits(provider, signer, { ipfsConfig: ipfsConfig }).init();
   }
 
+  static for (connectionOptions, kreditsOptions) {
+    const { network, rpcUrl, wallet } = connectionOptions;
+    if (!rpcUrl && network === 'local') { rpcUrl = 'http://localhost:8545'; }
+    let ethProvider, signer;
+    if (rpcUrl || network === 'local') {
+      ethProvider = new ethers.providers.JsonRpcProvider(rpcUrl);
+    } else {
+      ethProvider = new ethers.getDefaultProvider(network);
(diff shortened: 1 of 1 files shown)


lib/kredits.js
```diff
@@ -66,6 +66,23 @@ class Kredits {
...
+    if (rpcUrl || network === 'local') {
+      ethProvider = new ethers.providers.JsonRpcProvider(rpcUrl);
+    } else {
+      ethProvider = new ethers.getDefaultProvider(network);
+    }
+    if (wallet) {
+      signer = wallet.connect(ethProvider);
+    } else if (ethProvider.getSigner) {
+      signer = ethProvider.getSigner();
+   }
```
skddc: I think the linter should have caught the indentation here. It's off by one char.

lib/kredits.js
```diff
@@ -66,6 +66,23 @@ class Kredits {
     return new Kredits(provider, signer, { ipfsConfig: ipfsConfig }).init();
   }
 
+  static for (connectionOptions, kreditsOptions) {
+    let { network, rpcUrl, wallet } = connectionOptions;
+    if (!rpcUrl && network === 'local') { rpcUrl = 'http://localhost:8545'; }
+    let ethProvider, signer;
+    if (rpcUrl || network === 'local') {
```
fsmanuel: ` || network === 'local'` braucht man glaube ich nicht

PR #129, opened 2019-05-14
Add pre-commit hook and setup script

skddc:
Runs the appropriate linter on staged files before committing.

Can be easily installed in the local repo by running `npm run setup`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,11 +25,6 @@ Each of the aragon apps are separate packages:
 
 You can use `npm run install-all` to install all app dependencies at once.
 
-We also recommend using our git hooks for linting staged files and other
-convenient helpers. You can install them via:
-
-    $ npm run setup
-
 ### Local development chain
 
 For local development it is recommended to use
--- a/README.md
+++ b/README.md
@@ -25,6 +25,11 @@ Each of the aragon apps are separate packages:
 
 You can use `npm run install-all` to install all app dependencies at once.
 
+We also recommend using our git hooks for linting staged files and other
+convenient helpers. You can install them via:
+
+    $ npm run setup
+
 ### Local development chain
 
 For local development it is recommended to use
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   },
   "scripts": {
     "install-all": "./scripts/every-app.sh \"npm install\"",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -23,7 +23,8 @@
     "lint:contracts": "solhint \"contracts/**/*.sol\" \"apps/*/contracts/**/*.sol\"",
     "lint:contract-tests": "eslint apps/*/test",
     "lint:wrapper": "eslint lib/",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "setup": "sh scripts/git-hooks/install"
```
bumi:
```suggestion
    "setup-git-hooks": "sh scripts/git-hooks/install"
```

I find this name a bit confusing as it sounds it has something todo with the setup to install/run the code. We have `bootstrap`, `reset`, `install-all` - so I'd by a bit more specific here as so far this setup task is only installing the git hooks.

README.md
```diff
@@ -25,6 +25,11 @@ Each of the aragon apps are separate packages:
 
 You can use `npm run install-all` to install all app dependencies at once.
 
+We also recommend using our git hooks for linting staged files and other
+convenient helpers. You can install them via:
+
+    $ npm run setup
```
bumi:
```suggestion
    $ npm run setup-git-hooks
```

package.json
skddc: This setup is meant to include anything that is necessary, other than `npm install` to set up the repo itself.

skddc: Moved the entire thing to `postinstall` now.

skddc:
There's a weird issue with npm, in that when running a `postinstall` script it re-validates the entire npm dependency stack. I don't know why that is.

skddc:
@bumi Can we merge this now?

bumi:
Isn't `postinstall` also executed when this is installed as a dependency? e.g. in kredits-web? 

if it only runs when running `npm install` I am ok with merging it. though as much as I need this am not a fan of postinstall scripts doing such modifications. 
when we tried to run the app installations in a post-install script it was also only causing problems (don't remember the details though).

skddc:
Renamed it to your suggestion.

PR #130, opened 2019-05-17
Add JS wrapper for Kit contract

bumi:
This makes it possible to use a wrapper to create new DAOs.
At the same time this is to change the deploy script to be less dependent on truffle.


bumi:
can we merge this one?

skddc:
Didn't test it myself, but sure. LGTM.

PR #134, opened 2019-06-08
Update dependencies

bumi:


skddc:
There's an issue with Travis (not due to this change), and I'm not sure why it's using Yarn in the first place.

bumi:
hmmm... and what is `del-cli` https://www.npmjs.com/package/del-cli ?!

bumi:
seems it is caused by some postinstall script of a package: https://github.com/aragon/aragon-cli/blob/master/packages/aragon-cli/package.json#L11-L12 (actually scary that it does something like 'del-cli' on postinstall)

bumi:
using npm install it seems to work. @fsmanuel do you maybe have an idea what could be the issue here?

skddc:
Yay!

PR #135, opened 2019-06-11
Expose available network IDs

bumi:
This exposes the network IDs on which kredits is deployed. Helpful to
check if the used network is supported.


skddc:
thumbs-up-cool2

PR #139, opened 2019-06-13
Update aragon CLI to latest version

bumi:
updated aragon-cli everywhere and ran a fresh npm install everywhere. 
(removed node_modules and package-lock.json before installing)


skddc:
Unfortunately, the build is still red. :/

skddc:
Update: I get exactly the same error when `npm install`ing from this branch with empty `node_modules`.

skddc:
There seems to be some kind of chicken-and-egg problem. It only works when deleting the package lockfile beforehand. But a bootstrapped install then takes forever, due to the gazillion deps including native extensions.

bumi:
It seems to me that a `postinstall` script defined in the package.json of the aragon-cli is doing something: https://github.com/aragon/aragon-cli/blob/master/packages/aragon-cli/scripts/fix-lockfile

fund also this related: 
https://github.com/aragon/aragon-cli/blob/master/docs-internal/Dependencies.md#regenerate-the-lockfiles

bumi:
closing in favor of #140 and #142

PR #140, opened 2019-06-13
Update aragon CLI to latest version

bumi:
rebased version of #139


bumi:
It seems npm has some issues and a broken package-lock.json gets generated. 
`scripts/fix-package-lock.js` (which I have taken from aragon-cli) fixes the package-lock.json. 
This PR contains a fixed package-lock.json, so everything should work. But if new dependencies are added we probably have to run `node scripts/fix-package-lock.js` and be generally careful with changes on the package.json and package-lock.json

bumi:
This now also solves the apm problem. Now `open.aragonpm.eth` is also available on the devchain, so we can default to that one and no longer need to configure it depending on the network.

skddc:
I pulled the branch and ran `npm install`. Now I have a dirty package lockfile which seems to be from exactly those changes in the postinstall, but also includes changes to `ethereum-util`.

@bumi Could you please rebase those 5 commits before the last one and also make certain that you add the very last package lockfile in the end?

bumi:
this was my very latest npm files. what is the diff of yours?
seems subsequent `npm install` changes and breaks the lock file.  

There was a bug in the fix-package-lock script which used the wrong path (see last commit). 
Do you still have a dirty lock file after you ran the script?

skddc:
```diff
diff --git a/package-lock.json b/package-lock.json
index b40451a..a8971a8 100644
--- a/package-lock.json
+++ b/package-lock.json
@@ -6486,7 +6486,7 @@
           "resolved": "https://registry.npmjs.org/eth-block-tracker/-/eth-block-tracker-2.3.1.tgz",
           "integrity": "sha512-NamWuMBIl8kmkJFVj8WzGatySTzQPQag4Xr677yFxdVtIxACFbL/dQowk0MzEqIKk93U1TwY3MjVU6mOcwZnKA==",
           "requires": {
-            "async-eventemitter": "github:ahultgren/async-eventemitter#fa06e39e56786ba541c180061dbf2c0a5bbf951c",
+            "async-eventemitter": "async-eventemitter@github:ahultgren/async-eventemitter#fa06e39e56786ba541c180061dbf2c0a5bbf951c",
             "eth-query": "^2.1.0",
             "ethereumjs-tx": "^1.3.3",
             "ethereumjs-util": "^5.1.3",
@@ -6497,8 +6497,8 @@
           },
           "dependencies": {
             "async-eventemitter": {
-              "version": "0.2.3",
-              "resolved": "github:ahultgren/async-eventemitter#fa06e39e56786ba541c180061dbf2c0a5bbf951c",
+              "version": "github:ahultgren/async-eventemitter#fa06e39e56786ba541c180061dbf2c0a5bbf951c",
+              "from": "async-eventemitter@github:ahultgren/async-eventemitter#fa06e39e56786ba541c180061dbf2c0a5bbf951c",
               "requires": {
                 "async": "^2.4.0"
               }
@@ -10391,7 +10391,7 @@
                   "integrity": "sha512-NamWuMBIl8kmkJFVj8WzGatySTzQPQag4Xr677yFxdVtIxACFbL/dQowk0MzEqIKk93U1TwY3MjVU6mOcwZnKA==",
                   "dev": true,
                   "requires": {
-                    "async-eventemitter": "github:ahultgren/async-eventemitter#fa06e39e56786ba541c180061dbf2c0a5bbf951c",
+                    "async-eventemitter": "async-eventemitter@github:ahultgren/async-eventemitter#fa06e39e56786ba541c180061dbf2c0a5bbf951c",
                     "eth-query": "^2.1.0",
                     "ethereumjs-tx": "^1.3.3",
                     "ethereumjs-util": "^5.1.3",
@@ -10402,8 +10402,8 @@
                   },
                   "dependencies": {
                     "async-eventemitter": {
-                      "version": "0.2.3",
-                      "resolved": "github:ahultgren/async-eventemitter#fa06e39e56786ba541c180061dbf2c0a5bbf951c",
+                      "version": "github:ahultgren/async-eventemitter#fa06e39e56786ba541c180061dbf2c0a5bbf951c",
+                      "from": "async-eventemitter@github:ahultgren/async-eventemitter#fa06e39e56786ba541c180061dbf2c0a5bbf951c",
                       "dev": true,
                       "requires": {
                         "async": "^2.4.0"
@@ -27767,7 +27767,7 @@
       "integrity": "sha512-NamWuMBIl8kmkJFVj8WzGatySTzQPQag4Xr677yFxdVtIxACFbL/dQowk0MzEqIKk93U1TwY3MjVU6mOcwZnKA==",
       "dev": true,
       "requires": {
-        "async-eventemitter": "github:ahultgren/async-eventemitter#fa06e39e56786ba541c180061dbf2c0a5bbf951c",
+        "async-eventemitter": "async-eventemitter@github:ahultgren/async-eventemitter#fa06e39e56786ba541c180061dbf2c0a5bbf951c",
         "eth-query": "^2.1.0",
         "ethereumjs-tx": "^1.3.3",
         "ethereumjs-util": "^5.1.3",
@@ -27778,8 +27778,8 @@
       },
       "dependencies": {
         "async-eventemitter": {
-          "version": "0.2.3",
-          "resolved": "github:ahultgren/async-eventemitter#fa06e39e56786ba541c180061dbf2c0a5bbf951c",
+          "version": "github:ahultgren/async-eventemitter#fa06e39e56786ba541c180061dbf2c0a5bbf951c",
+          "from": "async-eventemitter@github:ahultgren/async-eventemitter#fa06e39e56786ba541c180061dbf2c0a5bbf951c",
           "dev": true,
           "requires": {
             "async": "^2.4.0"
@@ -31743,4 +31743,4 @@
       }
     }
   }
-}
\ No newline at end of file
+}
```

skddc:
After running the script manually, the tree is clean. I checked, and it is indeed missing the `postinstall` property in `package.json`.

bumi:
should we run it as `postinstall`? I am scared of automatic postinstalls and afik it will also run when used as dependency.

skddc:
Obviously that's not the case, because otherwise the `aragon-cli` postinstall script would fix our lockfile. If we don't run it automatically, then everybody using this repo always has this problem.

bumi:
no, the aragon-cli script would look for a different file: https://github.com/aragon/aragon-cli/blob/master/packages/aragon-cli/scripts/fix-lockfile#L23

also only people who would want to make changes to the package-lock.json (change dependencies) would need to run that script. Even though package-lock gets dirty this branch here works (also on travis)

skddc:
> also only people who would want to make changes to the package-lock.json (change dependencies) would need to run that script.

I didn't change anything about the packages and still had a dirty tree after the npm install. If nothing changed it should be clean after running install. It changing to something that is invalid/breaking is bad for developer UX and would inevitably result in someone committing the wrong lockfile from time to time as well.

skddc:
It's possible to know if the `postinstall` is run for a dependency or not. From what I read you can essentially just check if the current dir looks like your own package or not...

skddc:
I just tried using this branch as a dependency in `kredits-web` and the package from the lockfile fix doesn't appear in `kredits-web`'s own lockfile after `npm install`. So I think it shouldn't be an issue.

bumi:
aragon-cli has nothing to do with this anymore - I was wrong about aragon-cli messing something up. 

I guess the only question that remains is if we want to run such a hack on postinstall (which means it is run everytime kredits-contracts is installed as a dependency and not only when someone runs npm install within this repository.  

It for sure is a bad developer experience if a simple install gives you a dirty lock file. Though it seems that is some npm issue... I would like to understand that and actually fix the root cause instead of running some code that fixes this symptom. 
So my hope would be that this is something very temporary. (needing such a script is like wtf.)

Because I somehow dislike postinstall magic generally and because it would run everytime this is installed as dependency I'd only add a bold note in the readme. 
But I am also ok with adding it as postinstall if you want. 

Generally I hope we get rid of that soon™.

bumi:
aragon-cli has nothing to do with this anymore - I was wrong about aragon-cli messing something up. 

I guess the only question that remains is if we want to run such a hack on postinstall (which means it is run everytime kredits-contracts is installed as a dependency and not only when someone runs npm install within this repository.  

It for sure is a bad developer experience if a simple install gives you a dirty lock file. Though it seems that is some npm issue... I would like to understand that and actually fix the root cause instead of running some code that fixes this symptom. 
So my hope would be that this is something very temporary. (needing such a script is like wtf.)

Because I somehow dislike postinstall magic generally and because it would run everytime this is installed as dependency I'd only add a bold note in the readme. 
But I am also ok with adding it as postinstall if you want. 

Generally I hope we get rid of that soon™.

bumi:
as far as I understand it, npm install produces a broken lock file.
why? no idea, and also not how to fix it.
a readme entry + a script that fixes it is imo enough until the root cause is fixed.

but you did not really answer my question about the postinstall script. you vote for the postinstall?

skddc:
>  So my hope would be that this is something very temporary. (needing such a script is like wtf.)

Yes, hence my suggestion to ask the aragon-cli devs about it, because it looks to me like they also introduced it as a temporary solution.

> a readme entry + a script that fixes it is imo enough until the root cause is fixed. but you did not really answer my question about the postinstall script. you vote for the postinstall?

Yes, I vote for postinstall, because we have no idea when the root cause will be fixed, but we have plenty of people who should be able to run postinstall without having a broken lockfile as dirty tree after every install.

skddc:
I added a `postinstall` script which runs the fix and ignores any kind of errors (for any cases it may be run in, that could cause it to break).

However, it isn't actually run after the install for me!? My output looks like this:

```
(chore/update-dependencies-1 *) ~/src/kosmos/kredits/kredits-contracts $ npm install --no-audit                                                                              
                                                                                                                                                                             
> kredits-contracts@5.3.0 postinstall /home/basti/src/kosmos/kredits/kredits-contracts                                                                                       
> node scripts/fix-package-lock.js &>/dev/null || true                                                                                                                       
                                                                                                                                                                             
npm WARN acorn-jsx@5.0.1 requires a peer of acorn@^6.0.0 but none is installed. You must install peer dependencies yourself.                                                 
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.7 (node_modules/@aragon/cli/node_modules/ganache-core/node_modules/fsevents):                                   
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.7: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})            
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.9 (node_modules/@aragon/cli/node_modules/fsevents):                                                             
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.9: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})            
                                                                                                                                                                             
up to date in 16.835s          
```

And then the lockfile is dirty again. When I then run `npm run postinstall` manually afterwards, the lockfile is clean again. @bumi could you see what it does on your machine?

bumi:
yes, seems the package-lock is written after that script is executed? when I run `git status` in the postinstall I still have a clean master.

skddc:
Turns out it's a known issue since 2017: https://github.com/npm/npm/issues/18798

Solution is to use `postshrinkwrap` (a remnant of the shrinkwrap command which at some point was added to `install` by default). Now it works.

@bumi I think it's good to merge now. But could you squash/rebase your testing commits in this branch, please?

bumi:
closing in favor of #142

PR #141, opened 2019-06-17
Test setup

haythem96:
This PR setup test suite


haythem96:
should we change the port for the development network to `8545` as aragon ganache script use that port? or just not use that script ?

bumi:
we intentionally used a different port than the production RPC port. 
is this the ganache-cli.sh script? https://github.com/aragon/aragon-apps/blob/cb9b20c144a5670c8ee3e4bf1f0898897a8de35f/shared/test-helpers/ganache-cli.sh#L17

so the port is configurable as environment variable? and defaults to 8555?

haythem96:
@bumi cool! didn't check that the port is configurable...
the default port is 8545 https://github.com/aragon/aragon-apps/blob/cb9b20c144a5670c8ee3e4bf1f0898897a8de35f/shared/test-helpers/ganache-cli.sh#L24 as 8555 when running coverage

haythem96:
can we merge this one ?

bumi:
yes, looks good to me. sadly we have that merge conflict in the package-lock files because of the recent aragon update. maybe you can rebase this on master again?

PR #143, opened 2019-06-26
Contributor app tests

haythem96:
This PR add unit tests for the Contributor app contracts.


haythem96:
can we merge this one ?

apps/contributor/test/contributor.js
```diff
@@ -0,0 +1,170 @@
+const namehash = require('eth-ens-namehash').hash;
```
bumi: we should be able to use the namehash from ethers:
`const namehash = require('ethers').utils.namehash;`
as ethers.js is already a dependency we do not need to add another dependency

apps/contributor/contracts/Contributor.sol
```diff
@@ -64,10 +64,13 @@ contract Contributor is AragonApp {
   }
 
   function updateContributorAccount(uint32 id, address oldAccount, address newAccount) public auth(MANAGE_CONTRIBUTORS_ROLE) {
+    require(newAccount != address(0), "invalid new account address");
+    require(getContributorAddressById(id) == oldAccount, "contributor does not exist");
```
bumi: :+1: very good!

skddc:
Are these test files linted? The formatting (e.g. indentation) seems to be all over the place from what I can see in the GitHub diff...

apps/contributor/test/contributor.js
haythem96: cool will change that

haythem96: for the ganache script, it is basically the same aragon one, but I change 
```
run_tests() {
  echo "Running tests..."
  npx truffle test --network rpc "$@"
}
```
to 
```
run_tests() {
  echo "Running tests..."
  npx aragon contracts test --network rpc "$@"
}
```
Because `npx truffle test --network rpc "$@"` didn't work for me for no reason...

bumi:
ah ja, @haythem96 we used two spaces indentation on kredits. and you can run the linting using: `npm run lint:contract-tests` (though I had issues running the linter here right now. @skddc can you confirm this is correct?)

skddc:
> @skddc can you confirm this is correct?)

Confirm that what exactly is correct?

bumi:
@skddc that `npm run lint:contract-tests` is the right command to lint those files and that it works for you. 
on my machine it does not give me any output.

haythem96:
> @skddc that `npm run lint:contract-tests` is the right command to lint those files and that it works for you.
> on my machine it does not give me any output.

I tried it also, no output...

fsmanuel:
@haythem96 we don't have an explicit indentation rule, yet.
You can add `'indent': ['error', 2]` to the root `.eslintrc.js` in the rules section.

skddc:
>  that `npm run lint:contract-tests` is the right command to lint those files

I didn't create those commands and I never tried linting contract test files (no tests until very recently), so I don't know.

>  You can add `'indent': ['error', 2]` to the root `.eslintrc.js` in the rules section.

Sounds like a good idea to me. Thanks, @fsmanuel.

haythem96:
> @haythem96 we don't have an explicit indentation rule, yet.
> You can add `'indent': ['error', 2]` to the root `.eslintrc.js` in the rules section.

Cool, fixed now.

fsmanuel:
@haythem96 thanks for adding the rule. Now you got us:

```
lib/serializers/contributor.js
  16:1  error  Expected indentation of 2 spaces but found 1  indent
```

Can you please fix it! ❤️

haythem96:
> @haythem96 thanks for adding the rule. Now you got us:
> 
> ```
> lib/serializers/contributor.js
>   16:1  error  Expected indentation of 2 spaces but found 1  indent
> ```
> 
> Can you please fix it!

done.

skddc:
@haythem96 Did @bumi tell you about the kredits labels for issues and pull requests? This one is still missing the label...

haythem96:
> @haythem96 Did @bumi tell you about the kredits labels for issues and pull requests? This one is still missing the label...
> 
> (Documenting how this works should be solved during the current milestone, which is about onboarding.)

Yeah I know, I can't add a label...

skddc:
@bumi Could you make sure that @haythem96 is fully set up, please?

skddc:
The package lockfile is missing in this branch again. :/

skddc:
@haythem96 I opened a PR to your repo/branch, so the tests are run on Travis: https://github.com/haythem96/kredits-contracts/pull/1

bumi:
I've looked at the `ganache-cli.sh` script again and have a view questions/comments: 

* do we want to use npx there? all executables (aragon/truffle/ganache-cli) should be available as npm dependency anyway. 
* can we use `aragon --devchain` instead of running `ganache-cli` directly? 
* we don't have the `rpc` environment - we should probably add a `test` environment in the truffle.js

skddc:
Looks like Travis can't deal with merging PRs in repo forks: https://travis-ci.org/67P/kredits-contracts/pull_requests

I guess pushing one more commit should do the trick.

skddc:
If the other tests aren't supposed to work yet, then let's just change their npm scripts so that they don't run in `npm test` yet (e.g. change `test:token` to something like `echo 'WIP'` or similar).

haythem96:
> I just pulled this branch and ran `npm run test:contributor`, which I assume is what runs the code in here.
> 
> Unfortunately, it just hangs with no output, after printing "Passing the command to Truffle":
> 
> Screenshot from 2019-07-10 10-54-24

For some reasons if we change the env PORT variable and the post of the development network to `8545` it work...!

haythem96:
> If the other tests aren't supposed to work yet, then let's just change their npm scripts so that they don't run in `npm test` yet (e.g. change `test:token` to something like `echo 'WIP'` or similar).

just modified the npm test scripts to use the the ganache-cli script in the project instead of the one in @aragon/test-helpers so the other tests should work fine also...

haythem96:
> I've looked at the `ganache-cli.sh` script again and have a view questions/comments:
> 
> * do we want to use npx there? all executables (aragon/truffle/ganache-cli) should be available as npm dependency anyway.
> * can we use `aragon --devchain` instead of running `ganache-cli` directly?
> * we don't have the `rpc` environment - we should probably add a `test` environment in the truffle.js

So should we drop that ganache-cli script and use the devchain ?

bumi:
@haythem96 no idea :) I was just wondering. As `aragon devchain` is also using ganache-cli I thought that might be closer to the dev environment.

haythem96:
the tests should work now with running devchain manually before `npm run test`, 
didn't know how to config travis this way XD ...

haythem96:
I want more Kredits token :p  can we fix the travis config ? the tests work now when running the `devchain`

skddc:
What exactly is unclear about doing the same on Travis as you do locally?

These docs might clear things up: https://docs.travis-ci.com/user/job-lifecycle/

haythem96:
I think Travis build works now

apps/contributor/package.json
```diff
@@ -7,10 +7,10 @@
     "@aragon/cli": "^5.9.6"
   },
   "devDependencies": {
-    "@aragon/test-helpers": "^1.1.0",
     "eth-gas-reporter": "^0.2.0",
     "ganache-cli": "^6.4.3",
-    "solidity-coverage": "^0.5.11"
+    "solidity-coverage": "^0.5.11",
+    "@aragon/test-helpers": "^1.0.0"
```
bumi: do we still need those? and if so should we use the latest version?

package.json
```diff
@@ -44,6 +45,7 @@
     "@aragon/cli": "^5.9.6",
     "@aragon/kits-base": "^1.0.0",
     "@aragon/os": "^4.2.0",
+    "@aragon/test-helpers": "^1.0.0",
```
bumi: same here: do we still need this and if so should we use the latest version?

apps/contributor/package.json
haythem96: I used the test-helpers package in the tests. For the solidity-coverage we may need it... or should I remove it ?

bumi: no, I was only wondering about the test-helpers. can we use the latest version here? 2.0.0
https://www.npmjs.com/package/@aragon/test-helpers or does that beak things?

haythem96: I updated it now, did not break anything :)

skddc:
> I got two more comments and I think we are good to merge :shipit:
> what do you say? @skddc

I trust your opinion. But please clean up the git history when merging. ;)

bumi:
ok, cool. if we find issues or optimizations we do that in a new PR
I'd say we simply squasch merge this one.

PR #145, opened 2019-06-29
Vault app

haythem96:
Related to #169


bumi:
looks like a good start :+1: 
we should directly remove the files that we don't use, e.g. the html and frontend JS.

apps/vault/contracts/Vault.sol
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  function withdraw(address _token) external payable isInitialized {
+    uint256 contributorFundPercentage = (balanceOf(msg.sender) * 100) / totalSupply();
+    //to check if sender is a contributor
+    require(contributorFundPercentage > 0, "Contributor have no fund");
+
+    uint256 contributorFund = (address(this).balance * contributorFundPercentage) / 100;
+    msg.sender.transfer(contributorFund);
+
```
haythem96: @bumi we should burn the tokens here for the contributor that withdraw, right ?

PR #155, opened 2019-07-18
Fix send-funds script to use new web3 API

bumi:
The API changed with the recent aragon/truffle update.


skddc:
LGTM, but missing kredits label.

Issue #103, opened 2019-04-16
Testing infrastructure


fsmanuel:
We need to setup a testing infrastructure for solidity and the wrapper.

TODO:
- [x] Code linting https://github.com/67P/kredits-contracts/pull/97
- [ ] Wrapper contracts
- [ ] Serializer https://github.com/67P/kredits-web/pull/103
- [ ] Solidity contracts

bumi:
maybe @haythem96 has some input or can help with the solidity tests?

haythem96:
sure, will work on this those days...

skddc:
I think linting is there now, right?

haythem96:
> I think linting is there now, right?

Yes

PR #157, opened 2019-07-29
Test token app

haythem96:
This PR add unit tests for the Token smart contracts. #103

### DIFF
--- a/apps/token/contracts/Token.sol
+++ b/apps/token/contracts/Token.sol
@@ -21,6 +21,8 @@ contract Token is ERC20Token, AragonApp {
   }
 
   function mintFor(address contributorAccount, uint256 amount, uint32 contributionId) public isInitialized auth(MINT_TOKEN_ROLE) {
+    require(amount > 0, "amount should be greater than zero");
+
     uint256 amountInWei = amount.mul(1 ether);
     _mint(contributorAccount, amountInWei);
     emit LogMint(contributorAccount, amount, contributionId);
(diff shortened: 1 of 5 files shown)


haythem96:
can we merge this one ?

apps/token/contracts/Token.sol
```diff
@@ -21,6 +21,8 @@ contract Token is ERC20Token, AragonApp {
   }
 
   function mintFor(address contributorAccount, uint256 amount, uint32 contributionId) public isInitialized auth(MINT_TOKEN_ROLE) {
+    require(amount > 0, "amount should be greater than zero");
```
bumi: I think so far I always used keys like `INVALID_AMOUNT` instead of full messages. 
This should make the code smaller and is enough to identify the error and potentially also map it to a full message

apps/token/contracts/Token.sol
```diff
@@ -21,6 +23,8 @@ contract Token is ERC20Token, AragonApp {
   }
 
   function mintFor(address contributorAccount, uint256 amount, uint32 contributionId) public isInitialized auth(MINT_TOKEN_ROLE) {
+    require(amount > 0, ERROR_INVALID_AMOUNT);
+
```
haythem96: @bumi is that cool ?

bumi: just saw that I haven't done it everywhere. I thought about something like this: https://github.com/67P/kredits-contracts/blob/master/apps/contribution/contracts/Contribution.sol#L181

as that full string increases the contract and thus the deployment costs, afaik? so I thought it is better to use small strings.

haythem96: ahh okaay (y)

PR #159, opened 2019-08-02
WIP: Contribution contracts tests

haythem96:
Related to #103


apps/contribution/test/contribution.js
```diff
@@ -0,0 +1,231 @@
...
+      let contributionId = await contribution.contributionsCount();
+      return assertRevert(async () => {
+        await contribution.claim(contributionId.toNumber()+1, {from: root});
+        'contribution not found';
+      });
+    });
+
+    it("claim contribution", async () => {
+      let contributionId = await contribution.contributionsCount();
+      await contribution.claim(contributionId);
```
haythem96: this one revert for some reasons

apps/contribution/test/contribution.js
```diff
@@ -0,0 +1,231 @@
...
+    });
+
+    it("should revert when add contribution with amount equal to zero", async () => {
+      return assertRevert(async () => {
+        await contribution.add(0, contributorId, hashDigest, hashFunction, hashSize, {from: root});
+        'amount equal to zero';
+      });
+    });
+
+    it("add contribution", async () => {
```
skddc: Looks like the expectation wasn't described here (`it("does x")`). It also looks like there are 3 different expectations in one block here.

apps/contribution/test/contribution.js
```diff
@@ -0,0 +1,231 @@
...
+      let contributionCountBefore = await contribution.contributionsCount();
+      await contribution.add(amount, contributorId, hashDigest, hashFunction, hashSize, {from: root});
+      let contributionCountAfter = await contribution.contributionsCount();
+      // eslint-disable-next-line no-undef
+      assert.equal(contributionCountAfter.toNumber()-contributionCountBefore.toNumber(), 1);
+      let contributionObject = await contribution.getContribution(contributionCountAfter.toNumber());
+      // eslint-disable-next-line no-undef
+      assert.equal(contributionObject[1], contributorId);
```
skddc: I think you can also add expectation descriptions as the last argument to `assert()`, in case you really want to assert multiple things in one `it` block.

apps/contribution/test/contribution.js
skddc: Reverts unexpectedly? And in which context/scenario?

haythem96: in the scenario of claiming an added contribution

apps/contribution/test/contribution.js
haythem96: most of those multiple assert check for the same thing, that the contribution is added.. XD

apps/contribution/test/contribution.js
skddc: How would that work? You can only claim confirmed contributions.

apps/contribution/test/contribution.js
skddc: They very line I commented on asserts that the contributorId is correct, not that the contribution itself was added.

apps/contribution/test/contribution.js
haythem96: ahhhh yup okay, missed this ` require(block.number >= c.confirmedAtBlock, 'NOT_CLAIMABLE'); ` thanks!

apps/contribution/test/contribution.js
haythem96: yeah true, honestly don't know if that all those asserts are necessary... maybe just for more code coverage

skddc: In this case it's only important that you describe what you mean to test exactly. Currently, there's no description of what is being tested in this block.

apps/contribution/test/contribution.js
```diff
@@ -0,0 +1,294 @@
...
+
+      if(contributionId < 10) {
+        await timeTravel(100);
+      }
+      else {
+        await timeTravel(blocksToWait);
+      }
+      await mineBlock();
+
+      await contribution.claim(contributionId);
```
haythem96: this should fix the claim tests, no @skddc @bumi ?

apps/contribution/test/contribution.js
```diff
@@ -0,0 +1,349 @@
...
+      return assertRevert(async () => {
+        await contribution.claim(contributionId.toNumber(), {from: root});
+        'contribution already claimed';
+      });
+    });
+
+  });
+
+  describe("Veto claimed contribution", async () => {
+    // eslint-disable-next-line no-undef
```
haythem96: idk if this test is really necessary as if a contribution is claimed, so theoretically it is impossible to veto because of block number @bumi

apps/contribution/test/contribution.js
```diff
@@ -0,0 +1,349 @@
...
+
+    it("claim contribution", async () => {
+      if(contributionId < 10) {
+        await timeTravel(100);
+      }
+      else {
+        await timeTravel(blocksToWait);
+      }
+      await mineBlock();
+
```
haythem96: can't figure out why this one is not working, idk what I'm missing there! @bumi @skddc

apps/contribution/test/contribution.js
```diff
@@ -0,0 +1,349 @@
...
+    });
+
+    it("veto contribution", async () => {
+      const contributionId = await contribution.contributionsCount();
+      let contributionObject = await contribution.getContribution(contributionId.toNumber());
+      console.log("veto block: " + contributionObject[7]);
+      console.log("current block: " + await getBlockNumber());
+
+      await contribution.veto(contributionId.toNumber(), {from: root});
+      // eslint-disable-next-line no-undef
```
haythem96: blocks get minted and the veto period end right away... @bumi  @skddc

apps/contribution/test/contribution.js
```diff
@@ -45,6 +45,16 @@ const mineBlock = function() {
   });
 };
 
+const getBlockNumber = function() {
```
skddc: getBlockNumber should actually return a promise as well, so this entire function is superfluous. We're using it with a promise e.g. here: https://github.com/67P/kredits-web/blob/master/app/services/kredits.js#L150

haythem96: it return also there a promise, no ?

skddc: The function you added is just wrapping the normal callback in a promise. Wherever it's called you can just use the original function with a promise.

haythem96: I updated it, the thing is blocks get mined instantly so veto period end...

apps/contribution/test/contribution.js
skddc: Yes, the first 10 are confirmed immediately iirc.

haythem96:
As an update: 
- still have one error with the claim contribution test.
- for travis, even when tests fail, it exit with 0

.travis.yml
```diff
@@ -26,7 +26,11 @@ before_script:
 script:
   - npm run lint:wrapper
   - npm run lint:contract-tests
-  - npm run test
+  - npm run test:token
+  - npm run test:contributor
+  - npm run test:contribution
+  - npm run test:proposal
+  - npm run test:kit
```
skddc: This should be the same as the test script itself, because `&&` is supposed to exit with 1, if any of the commands on the way exit with 1.

haythem96: yeah true, I changed that also to 
```env:
  - TASK=lint:wrapper
  - TASK=lint:contract-tests
  - TASK=test:token
  - TASK=test:contributor
  - TASK=test:contribution
  - TASK=test:proposal

before_script:
  - npm run devchain > /dev/null &
  - sleep 5

script:
  - travis_wait 60 npm run $TASK
```
It is the same also....
But the thing is there is an error in Contribution tests, but it doesn't exit with 1!

skddc: I don't see how the env vars make it clearer. I think it's actually less understandable that way. In fact, I wouldn't even know off the top of my head what happens when you assign the same env variable twice.

>  But the thing is there is an error in Contribution tests, but it doesn't exit with 1!

Yes, that's the thing that needs to be fixed. Not how the tests are run. The test task should just be `npm test` and nothing else.

haythem96: What I meant by the env vars make it clearer, in Travis u don't need to look at the whole log in case a test fail...

package.json
```diff
@@ -44,6 +44,7 @@
   "devDependencies": {
     "@aragon/kits-base": "^1.0.0",
     "@aragon/os": "^4.2.0",
+    "@aragon/test-helpers": "^2.0.0",
     "async-each-series": "^1.1.0",
```
haythem96: Adding the test-helpers package fixed #163

.travis.yml
skddc: Oh, I see. The feature you're looking for is the build matrix:

https://docs.travis-ci.com/user/customizing-the-build/#naming-jobs-within-matrices

haythem96:
as an update for this PR; build run well... there is a revert in the claim contribution test, but test doesn't exit with exit code 1...

this call revert:  `    address contributorAccount = getContributorAddressById(c.contributorId);`

When I try to call any function in the Contribution app that call another app function, it revert.
like `getContributorIdByAddress()`

Am I missing something in the tests ?

raucao:
Closing due to inactivity. Please feel free to re-open.

raucao:
Closing due to inactivity. Please feel free to re-open.

PR #160, opened 2019-08-04
Fix package script command deploy:kit

fsmanuel:


package.json
```diff
@@ -16,7 +16,7 @@
     "bootstrap": "npm run reset:hard && npm run seeds",
     "reset": "npm run deploy:kit && npm run deploy:dao",
     "reset:hard": "npm run deploy:apps && npm run reset",
-    "deploy:kit": "npm run compile-contracts && aragon contracts exec scripts/deploy-kit.js",
+    "deploy:kit": "aragon contracts exec scripts/deploy-kit.js",
```
bumi: what was actually broken? are the contracts still compiled?

skddc: The command fails both on @fsmanuel's an my machine. Compiling the contracts isn't the same as deploying something. In the bootstrap, the contracts are compiled before this step.

bumi: ok, but then we need to fix the `compile-contracts` command? 

at which step are the contracts compiled?

skddc: I guess so. Frankly, I don't know what is needed and what isn't, but this one was consistently failing and nobody fixed it. So this change is more of a hotfix to get the bootstrap to work, because most people don't change the kit contract or other such contracts.

bumi: is there an issue for that? `compile-contracts` works fine for me (just tested it again)
even though people won't change the kit, I assumed the kit still needs to be compiled to run the deploy script.

bumi: if you have a fresh clone (without the `build` directory) you'll now get a: 
`Error: ENOENT: no such file or directory, scandir '[...]/build/contracts'` error. 
so contracts need to be deployed, it probably works for you now because you had deployed contracts in the `build` directory. 

I can not reproduce any compile error here. :(

skddc: > I can not reproduce any compile error here. :(

Then we need to add the complete install and bootstrap on Travis, because @fsmanuel and I both have the same compilation errors on two different operating systems with different environments.

bumi: good idea. can you create an issue with the error you've seen? thanks

Issue #158, opened 2019-07-29
Don't mine blocks instantly on devchain


skddc:
In Ganache, it is possible to deactivate auto-mining and set a block time in seconds.

Background is that I'm actually stuck with Web UI development right now, because everything is instantly confirmed, but want to design exactly the situation where there's a waiting time in between action and confirmation (https://github.com/67P/kredits-web/issues/120).

I was also wondering why we don't use the actual Ganache app, which is available for all of Linux, Mac, and Windows, where one can inspect and configure all kinds of devchain details via its UI. Would that be possible with our setup maybe?

skddc:
By the way, the argument for ganache-cli would be `--blocktime`, but `aragon devchain` doesn't seem to allow most additional chain options.

skddc:
Asked on Aragon Chat, because the Internet didn't know. Let's see.

skddc:
Yay:

>  Hey @raucao I recently include some additional options to configure the devchain (e.g. gas-limit, network-id, mnemonic). I can include `--blocktime` and other relevant configurations. Do you have some additional suggestion on that?
>
> Note that this will be possible on the next aragonCLI version that we will publish in the next couple of days

skddc:
From last night:

> @raucao Now with last aragonCLI version you are able to configure `--blocktime` with `aragon devchain`. I'll also include the option on `aragon run` on next patch release. Documentation will be updated soon as well

PR #162, opened 2019-08-09
Remove Aragon CLI, document system deps

skddc:
In order to prevent regular issues with the notoriously npm-issue-ridden Aragon CLI module, this removes it as a direct dependency and requires it to be installed as a system dependency.

Also adds Truffle as a system dependency to the docs (which was missing before).

The latest Aragon CLI asks has a lot of prompts for every contract deployment. I'll see if we can remove those by specifying more arguments upfront.

refs #158


bumi:
this is awesome! thanks for looking into it. 
maybe we can add the used/tested aragon and truffle version to the readme? 

and I think that's a `kredits-2`

skddc:
>  maybe we can add the used/tested aragon and truffle version to the readme?

Then we have to update it like every two days. They release *a lot* over at Aragon. :)


> and I think that's a `kredits-2`

Nah, 1 is perfectly fine. Didn't take long.

skddc:
Tests are also green after adding sys deps. Merging...

bumi:
> Then we have to update it like every two days. They release _a lot_ over at Aragon. :)

yep, but also breaking changes. - at least we have to keep in mind which versions we are using/supporting

skddc:
Whoever sees something breaking with the latest version on their system should open an issue imo.

PR #164, opened 2019-08-09
Remove cached artifacts

bumi:
Thos should be generated on every contract app deployment and thus don't need to
be in the repo.
We also don't have those for the other contract apps.

But I also don't know the details about those files....


skddc:
Oh god, these may be the reason for the last remaining prompts during bootstrap, when it finds artifact mismatches.

skddc:
Needs a kredits label...

bumi:
I just found those when trying a completely fresh bootstrap in #162

PR #166, opened 2019-08-09
Remove yarn lockfiles

skddc:
We have a bunch of security alerts, all from old versions in yarn lockfiles: https://github.com/67P/kredits-contracts/network/alerts

As only @fsmanuel was using Yarn in the past, I want to propose removing the lockfiles, so other people don't have to keep yarn support in mind when updating dependencies. Otherwise we actually have to duplicate everything we do with npm on Travis, in order to test it with yarn in addition to npm.

@fsmanuel What do you think?


bumi:
@fsmanuel is npm working for you now?

Issue #169, opened 2019-08-26
Create a vault app that can hold ETH for the DAO


bumi:
The DAO needs to be able to receive ETH ~~and other tokens (let's start with ETH only)~~ which will be used to distribute to token owners and to pay for expenses. 

For this we need a "vault" app that will handle the funds.

bumi:
@haythem96 already started with #145

haythem96:
moved the PR here #172

skddc:
Note from call: we agreed to use Aragon's vault app instead of our custom one. See #173.

PR #172, opened 2019-08-27
Vault app

haythem96:
Related to #169


bumi:
Let's try to re-used the default aragon vault app. if that works. we don't have to maintain our own. 
for that we need to "install" the aragon app to the DAO in the DAOkit and somehow enable it for existing DAOs (is there a default aragon way to do that?)

PR #175, opened 2019-09-17
Add support for zoom profiles

bumi:
This adds an accessor for the zoom_name to the contributor profile.
Doing this also removes general support for preserviing the contributor
accounts array as this was buggy and caused accounts to be added
multiple times.

### DIFF
--- a/lib/serializers/contributor.js
+++ b/lib/serializers/contributor.js
@@ -27,8 +27,7 @@ class Contributor {
       github_uid,
       github_username,
       gitea_username,
-      wiki_username,
-      accounts,
+      wiki_username
     } = this;
 
     let data = {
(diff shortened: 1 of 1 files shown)


bumi:
sadly I have issues with my setup since the weekend after I did a system upgrade and I don't get aragon cli and some native dependencies installed. so far I failed to fix it, thus this is untested PR so far.

skddc:
There are some Travis errors you can look at (and linting should work locally without Aragon CLI, too).

skddc:
> sadly I have issues with my setup since the weekend after I did a system upgrade

You don't have a way of rolling back your system to a working state? If you still use Arch, I can highly recommend using Timeshift snapshots before upgrades, so you don't get into this situation again.

skddc:
Tested successfully with my WIP branch of kredits-web. :+1:

bumi:
yeah, thanks for the hint. I am using timeshift, but it is some node.js and npm only issue I did not want to rollback.

PR #182, opened 2020-02-08
WIP: Fix Aragon breaking changes; update dependencies

### LINKED ISSUE #179: Compatibility with latest stable Aragon CLI
Running e.g. the devchain fails with "apm" missing from the apps' environment config:

```
/home/basti/.nvm/versions/node/v12.14.0/lib/node_modules/@aragon/cli/src/lib/environment/loadConfigFiles.js:37
    throw Error(
          ^
Error: Error parsing arapp.json:
arapp.environments['development'].apm should be object
arapp.environments['rinkeby'].apm should be object
arapp.environments['default'].apm should be object
```

skddc:
Dance the update dance with me.

### arapp.json exception

The only [documented options for the `apm` config property](https://hack.aragon.org/docs/cli-global-confg#the-arappjson-file) are now related to IPFS. So I removed what seem to be outdated values on our end, and the devchain works again.

### Other Aragon CLI changes

* The Aragon CLI package was still a dependency in every app's `package.json`. I removed it everywhere.
* There is no `aragon contracts` anymore. Neither `compile` nor `test`. It looks to me like `truffle test` should work (it runs stuff and complains about things), but there are test failures now. Probably due to other oudated API calls in the tests (e.g. `dao.acl()`).

### App package questions

@haythem96, @bumi: To be honest, the separate `package.json` files look like a mess to me, and I don't understand how these are used exactly, because the Travis config is not actually running `npm install-all` before running tests. Travis also doesn't seem to compile contracts before testing, so unless you commit and push compiled changes, I don't see how it could catch test failures in PRs.

These question should be solved either in this PR or with a new one focused on fixing the test setup. I won't touch the code in this branch from here on, so please feel free to work on this branch and push fixed tests.

closes #179


bumi:
Thanks for this PR! I could get the contracts running using this update. (pushed a minor update here)

I am not sure about travis and the tests currently fail on my machine: 

```
$ npm run test

1) Contract: Token app
       "before all" hook:
     TypeError: dao.acl is not a function
      at Context.before (test/token.js:25:43)
```

scripts/deploy-kit.js
```diff
@@ -13,7 +13,7 @@ const KreditsKit = artifacts.require('KreditsKit')
 
 const arapp = require('../arapp.json')
 const environment = argv['network'] || argv['environment'] || 'development'
-const apm = arapp.environments[environment].apm
+const apm = arapp.environments[environment].apm || 'open.aragonpm.eth'
```
skddc: `arapp.environments[environment].apm` has to be an object now, so the first condition here is not going to work anymore.

bumi: what does not work anymore? 

that `apm` entry in the arapp.json (removed in this PR) was added/used by us - and not aragon related. 
We had used that entry when `open.aragonpm.eth` was not available on all networks (and we had to use `aragonpm.eth` locally. 

now we can default to `open.aragonpm.eth`

bumi:
I've updated some tests, now it fails with: 

```
  1) Contract: Contributor app
       "before all" hook:
     Error: invalid bytes value (arg="_initializePayload", coderType="bytes", value=0)
     at PromiEvent (/home/bumi/.node_modules/lib/node_modules/truffle/build/webpack:/packages/contract/lib/promievent.js:9:1)
      at TruffleContract.newAppInstance (/home/bumi/.node_modules/lib/node_modules/truffle/build/webpack:/packages/contract/lib/execute.js:169:1)
      at Context.before (test/contributor.js:37:31)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

scripts/deploy-kit.js
skddc: Please read my original PR description and follow the link. The entire reason for opening this PR was an exception about this particular `arapp.json` property having to be an object now.

skddc: > that `apm` entry in the arapp.json (removed in this PR) was added/used by us - and not aragon related.

That would explain why it's not documented. They introduced this property themselves in the meantime, but with different keys.

bumi: this is a WIP PR, isn't it? 
I was trying to debug that issue and get this update running here.

that configuration conflict is not solved, yet and the tests do not run, yet.

re: `apm` entry in the arapp.json file I tend to not have the apm registry configurable anymore. We now can rely on `open.aragonpm.eth` (locally and for all Ethereum netwroks we want to deploy to)

skddc: Not sure how your comment relates to what I'm trying to explain. The config issue was solved by me removing the custom `apm` property, because Aragon now requires it to be their own documented object as a value (the one with the IPFS configs). The test failures are a different topic.

I explained these things in the original PR description, with headlines and separate paragraphs and all.

skddc: To summarize my one comment on this one line of code again:

`arapp.environments[environment].apm` cannot be a string with an APM address anymore, as per their documentation. And it fails with an exception when trying to execute any aragon-cli command that encounters this config property as a string.

I hadn't seen that it was used in other places, so what fixed the aragon commands for me was just to remove the keys from `arapp.json`.

bumi: which broke the setup/deployment of kredits and that's what I am trying to fix here... 
It seems you did not change any code there?!

And I know that `apm` is now a aragon related property and can not be used and I also read your comments... 

maybe I should better make a PR to your PR then we don't need to somewhat discuss WIPs and avoid missunderstandings.

bumi:
ToDo: 

- [ ] Update daoFactory address to use v0.8 of aragonOS. This should be the one deployed by the updated aragonOS package
see: https://github.com/aragon/deployments/tree/master/environments/rinkeby

scripts/deploy-kit.js
skddc: I cannot know what you know, because I don't have a telepathic link to you. Until this last comment it seemed like you hadn't understood my original comment. And my original comment was *merely a pointer* to that config option being outdated.

If you already understand something that someone points out, then why not just say "Yeah, I know, I'm working on that"? Or even nothing at all, if it's of no consequence to my involvement.

bumi: to me it seems you did not understand what that `apm` was used for and why your change broke our deploy script. 
but I also don't know what you know, because I also don't have a telepathic link to you.

anyhow this discussion now seems to be unrelated to this PR and also on already outdated code, so I am resolving this now.

bumi:
merging this as discussed. The other fixes should be done in separate PRs.

bumi:
btw. @haythem96 can you help with the contract upgrades?

PR #194, opened 2020-04-15
Some required dependency updates

bumi:
those are some dependency updates I had to add to run on the latest aragon version.


skddc:
Want to fix the linter error so we can merge this? Would be good to branch off from master with up-to-date dependencies...

skddc:
I just had another look at Travis, after fixing the linter issue, and it turns out the tests are being run before the devchain has finished starting up, due to it just being sent to background with `&`.

skddc:
After fixing the devchain startup issue, I found that the tests are failing both locally and on Travis, but with different errors.

Locally, there's an actual test failure in `token`, with an exception for a wrong argument. On Travis, the tests don't compile in the first place:

> Error: Cannot find module '@aragon/test-helpers/assertThrow'

I commented the contract tests in the Travis config, so they can be fixed outside of this PR.

PR #196, opened 2020-05-27
Chore/dependency updates

bumi:
Updated the npm dependencies to the latest versions

missing: 
- [ ] ipfs-http-client - which has some breaking changes.  I think it is not a problem for us but needs to be tested
- [ ] solhint - wich also has a breaking change and now needs pretier for codestyle linting. I did not want that additional dependency now. Also we probably needs to improve the linting of the solidity files

### DIFF
--- a/apps/contribution/package-lock.json
+++ b/apps/contribution/package-lock.json
@@ -5,22 +5,156 @@
   "requires": true,
   "dependencies": {
     "@aragon/os": {
-      "version": "4.3.0",
-      "resolved": "https://registry.npmjs.org/@aragon/os/-/os-4.3.0.tgz",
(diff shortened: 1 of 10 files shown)


bumi:
the new IPFS package now returns a `Object [AsyncGenerator] {} ` and no longer just a promise. 
we should update that package in an separate pull request. same for the solhint package.

I've tested all the scripts (bootstrap, etc.) and seems to work fine.

bumi:
those warnings: https://github.com/67P/kredits-contracts/network/alerts should be fine now, too. 
they are all from within the app/ package-lock.json

skddc:
utACK

Issue #195, opened 2020-05-07
Set up Reimbursement contract


raucao:
* [ ] Skeleton contract
* [ ] Test suite
* [ ] Tooling (whatever is necessary for Aragon app etc.)

Issue #48, opened 2018-05-24
Expense tracking & reimbursements


raucao:
This is a big one, and I think one GitHub issue is probably not the perfect venue for discussing it, but maybe we can hack a quick MVP for it first, then iterate on it and open seperate issues.

The first step here would be the pure tracking, not the actual remuneration. So, for example, @5apps is currently sponsoring servers for various open-source projects, including Kosmos. Ideally, one could create an expense of type "recurring/monthly" and just enter all monthly expenses one contributes combined, with the details in the description. Later on, we can then use that data, but still decide what and from which point in time to remunerate. I.e. we don't have to decide how to actually treat those expenses yet, but at least we can start tracking the pure data.

bumi:
I think it is easiest to do the recurring/monthly expensives off-chain first and create each month an expense on-chain. I think this makes it easier to do the actual payout and have a paid/unpaid status. 

I'd suggest storing the following data: 

* recipient (address) 
* amount
* token - ETH, wrapped bitcoin, whatever the vault supports)
* IPFS hash - description with details of the expense with e.g. invoices
* paid (bool) 
* approved (bool) - approved by voting for example

bumi:
The expense tracking app would call the vault (#172 #145) to do the payment. (the recipient can withdraw through the expense app)

skddc:
Somehow, we did not document in here what we discussed in the calls since then. The idea was to do it completely manual at first, and document what we do, so we can translate it into an on-chain solution.

We can actually start doing this now, because we have enough money for infrastructure sitting in a new Kosmos [multi-sig safe](https://gnosis-safe.io/app/#/safes/0x9CC29b8373FF92B01C1f09F31B5DD862350c167E/balances) since yesterday.

@bumi @haythem96 @velasquez @fsmanuel Anyone of you interested in joining a brainstorming/planning call about this?

bumi:
yeah, let's do a planning call. - for me ideally anytime next week. 

:+1: for a manual process first to see exactly what needs to be implemented on-chain. 
 
though I'd start with recording expenses on-chain soon. that should be rather simple as it just stores some data and references to an IPFS document. 
Did we already discuss what schema we want to use? anybody knows if there are some standard schemas?

skddc:
It's only Tuesday. I was actually hoping to manually log some expenses this week, and thus have some requirements ready for implementing on-chain. We need a veto window, as with everything else, so the payouts are going to be delayed.

[...]

PR #198, opened 2020-05-29
Add Reimbursement app

bumi:
refs #48, #195

## To do

* [x] Create reimbursement contract
* [x] Fix txs failing in kredits-web expenses branch
* [x] Rename `contributor` to `recipient`
* [ ] update kredits-web to use recipientId (instead of contributorId)


apps/reimbursement/contracts/Reimbursement.sol
```diff
@@ -0,0 +1,119 @@
...
+  function totalAmount(bool confirmedOnly) public view returns (uint256 amount) {
+    for (uint32 i = 1; i <= reimbursementsCount; i++) {
+      ReimbursementData memory r = reimbursements[i];
+      if (!r.vetoed && (block.number >= r.confirmedAtBlock || !confirmedOnly)) {
+        amount += r.amount; // should use safemath
+      }
+    }
+  }
+
+  function getReimbursement(uint32 reimbursementId) public view returns (uint32 id, address recipient, uint256 amount, address token, bool claimed, bytes32 hashDigest, uint8 hashFunction, uint8 hashSize, uint256 confirmedAtBlock, bool exists, bool vetoed) {
```
skddc, line 58: Maybe we should just call this `get`, analog to `add` below?

apps/reimbursement/contracts/Reimbursement.sol
```diff
@@ -0,0 +1,119 @@
...
+    ReimbursementData storage r = reimbursements[reimbursementId];
+    require(r.exists, 'NOT_FOUND');
+    require(!r.claimed, 'ALREADY_CLAIMED');
+    require(block.number < r.confirmedAtBlock, 'VETO_PERIOD_ENDED');
+    r.vetoed = true;
+
+    emit ReimbursementVetoed(reimbursementId, msg.sender);
+  }
+
+  function claim(uint32 reimbursementId) public isInitialized {
```
skddc, line 103: I think until we actually have the financial code done, we should not add this function or the `claimed` property. Both because it will actually be out of sync with what we actually send around from the Gnosis safe, and also because we may want to adopt a different approach here as well (probably similar to the contributions, and maybe even combined with dividend payouts, so you only pay for a single tx when withdrawing everything at once).

config/seeds.js
```diff
@@ -38,6 +38,8 @@ const contractCalls = [
   ['Contribution', 'add', [{ contributorId: 1, contributorIpfsHash: 'QmWKCYGr2rSf6abUPaTYqf98urvoZxGrb7dbspFZA6oyVF', date: '2019-04-11', amount: 5000, kind: 'dev', description: '[67P/kredits-contracts] Introduce contribution token', url: '' }, { gasLimit: 350000 }]],
   ['Contribution', 'add', [{ contributorId: 2, contributorIpfsHash: 'QmcHzEeAM26HV2zHTf5HnZrCtCtGdEccL5kUtDakAB7ozB', date: '2019-04-11', amount: 1500, kind: 'dev', description: '[67P/kredits-web] Reviewed stuff', url: '' }, { gasLimit: 350000 }]],
   ['Contribution', 'claim', [1, { gasLimit: 300000 }]],
+
+  ['Reimbursement', 'add', [{title: 'Server Hosting', description: 'All the serverz', amount: 1, token: '0xa3048576e296207eb0141f2803590ad044f81928', currency: 'WBTC', date: '2020-05-28'}, { gasLimit: 300000 }]],
```
skddc, line 42: This seems inconsistent with the expense schema. I don't see how it creates the array of expenses, separate from the WBTC reimbursement. Would also be good to add two expenses for an example reimbursement from the start (ideally just the same example as in the schemas repo) for UI development.

lib/addresses/KreditsKit.json
```diff
@@ -1,4 +1,5 @@
 {
   "4": "0x76e069b47b79442657eaf0555a32c6b16fa1b8b4",
+  "8653663": "0x0Dd2b0F3F8Cba588fa865abf289594e2cfE30995",
```
skddc, line 3: Are these supposed to be committed?

bumi: arg, noo. of course not.

config/seeds.js
bumi: the array here is not an array of expenses but just how it is passed to the `add` function. see above the seeds for contributions. 

The data here is not correct yet. and yep, we can add more just by c&p this line.

apps/reimbursement/contracts/Reimbursement.sol
bumi: yeah, thought about that, too. but we had `getContribution` in the contribution contract. 
do you think that's the best pattern? :+1:

skddc: Yeah, I think it should be consistent, and `contribution.add` is nicer for clients than `contribution.addContribution`. May as well imply the scope for all functions then.

config/seeds.js
skddc: > the array here is not an array of expenses but just how it is passed to the `add` function. see above the seeds for contributions.

Yes, that's not what I meant. I meant that a reimbursement is an array of expenses, so it needs be an array for the `add` function and the examples and seeds as well.

skddc:
@bumi Any news on this PR?

skddc:
I think I'm running into the same issue as Travis: the deployment artifacts seem to be undefined (according to the messages on Travis), resulting in the contract not being able to deployed during bootstrap on my machine:

Screenshot from 2020-06-13 11-42-00

lib/addresses/dao.json
```diff
@@ -1,3 +1,4 @@
 {
-  "4": "0xc34edf7d11b7f8433d597f0bb0697acdff55ef14"
+  "4": "0xc34edf7d11b7f8433d597f0bb0697acdff55ef14",
+  "69041181": "0x356685A0d9d66d6e5dA80f50EC206Af8009C8b6F"
```
bumi:
```suggestion
  "4": "0xc34edf7d11b7f8433d597f0bb0697acdff55ef14"
```

lib/addresses/KreditsKit.json
```diff
@@ -1,3 +1,4 @@
 {
-  "4": "0x76e069b47b79442657eaf0555a32c6b16fa1b8b4"
+  "4": "0x76e069b47b79442657eaf0555a32c6b16fa1b8b4",
+  "69041181": "0x6Ad8CDF71C3E2AaD2712964097b475184a0dfDeF"
```
bumi:
```suggestion
  "4": "0x76e069b47b79442657eaf0555a32c6b16fa1b8b4"
```

raucao:
I have updated the PR description with remaining tasks from the original issue as well as the Summit notes.

raucao:
67P/kredits-web/pull/195 is finished and ready for review/testing now. It will only need very small adjustments after finishing this PR.

bumi:
do we need to save who added the reimbursement? 
with the permissions it must be a contributor, but I had removed that we store who it is (mainly to avoid the lookup)

raucao:
> do we need to save who added the reimbursement?

If we have the tx sender available in clients already, then we wouldn't need an additional attribute I guess. We just need that piece information in general.

bumi:
I am not completely sure how to understand and explain the problem with deploying this aragon app. 
When I try to deploy this app locally and install it into our DAO it seems to work but somehow it does not show up when I do `aragon  dao apps ..` and the JS also can not find the contract address. 
So far I failed to debug this issue. 

Steps to do this:

1. `aragon apm publish major --propagate-content=false --build=false --prepublish=false --skip-confirmation`
2. `aragon dao install 0xbBc85e5dE36826e7fD52a871dB7e1735992e3AD4 kredits-reimbursement.open.aragonpm.eth`
3. `aragon dao apps 0xbBc85e5dE36826e7fD52a871dB7e1735992e3AD4`

I am not sure if that is related to #199

bumi:
> If we have the tx sender available in clients already, then we wouldn't need an additional attribute I guess. We just need that piece information in general. (Same for proposing kredits actually.)

I don't fully understand. what do you want to store? 

when adding an entry we have the `msg.sender` which is the address. to lookup the contributor for that address we need to somehow talk to the contributor contract. This so far has always been a pain and I am not sure how to best connect the contracts to each other.  (which is basically also the blocker still for deploying this)

raucao:
Why is deploying this preventing the PR from being merged? I think that should be done in a whole different PR, as that work is not about feature development for reimbursements.

raucao:
> I don't fully understand. what do you want to store?

I don't "want to store" anything that's not necessary. I simply mean that we need to be able to show who proposed a reimbursement later.

raucao:
I just tried doing the last `kredits-web` task for this, but as this branch is now, I get exceptions from ethers when trying `get` a reimbursement, no matter from where. I wasn't able to fix it so far. Seeds run OK, but then it complains about insufficient data for assigning a boolean value to `vetoed` when loading a reimbursement in JS.

raucao (approved):
Let's merge this, as it's working fine in dev, and we have to adapt everything for the production deployment and migration.

PR #199, opened 2020-06-13
Correct inter-contract associations

bumi:
So far our contract discovery was wrong.
https://spectrum.chat/aragon/aragonos/how-do-i-get-the-appid-namehash-of-an-app-in-open-aragonpm-eth~50dd2c67-63a2-49cf-bc64-aa033e33f48d

All dependent contract connections that are needed must be passed in and
can not be discovered from the DAO/Kernel/whatever.

So far I have no idea how we upgrade to these new contracts.


bumi:
actually not sure how to best proceed with such a change as I am not sure how that change would affect the deployed contracts or how to deploy those changes.

skddc:
Maybe it's easier to first solve bootstrapping/migrating contracts with existing data?

raucao:
Closing in favor of #213.

PR #200, opened 2020-06-24
Update to ethers 5

bumi:


skddc:
Sure that's a kredits-1 and not a 2?

bumi:
yeah no idea. was quite some trouble and work, but in the end it's not that much...

skddc:
Bootstrap and scripts work for me. But the hubot-kredits PR does not. See comment on 67P/hubot-kredits#59.

bumi:
merging this and preparing a new release.

PR #207, opened 2020-11-17
Lock node.js to 12 on CI, add note for devs to README

raucao:
Add note about having to use node.js 12 for now.


galfert:
Funnily enough, Travis fails for this PR because it's using Node 14 :)

bumi (approved):

PR #208, opened 2020-12-16
Add script to count contributions between a certain block timeframe

bumi:
I think it is broken so far, but maybe a start.

### DIFF
--- a/scripts/list-contributions-per-contributor.js
+++ b/scripts/list-contributions-per-contributor.js
@@ -25,16 +25,25 @@ module.exports = async function(callback) {
     let confirmedBeforeBlock = await promptly.prompt('Before block: ');
     let confirmedAfterBlock = await promptly.prompt('After block: ');
 
+    let tokens = {};
+    let contributors = await kredits.Contributor.all();
+    contributors.forEach(c => {
+      tokens[c.id] = { amount: 0, contributor: c };
+    });
+
     let contributionId = await kredits.Contribution.contract.contributionsCount();
     let nextContribution = true;
-
-    let contributors = {};
 
     while (nextContribution) {
       console.log(`Getting contribution: ${contributionId}`);
       let contribution = await kredits.Contribution.getById(contributionId);
       contributionId = contributionId - 1;
 
+      // if no conribution is found
+      if (!contribution.exists) {
+        nextContribution = false;
+        break;
+      }
       // check if the contribution is older
       // in that case we assume all other contributions now are older
       if (contribution.confirmedAtBlock < confirmedAfterBlock) {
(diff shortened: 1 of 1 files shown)


bumi:
Currently the numbers are not the same as on kredits-web. but I don't know why :)

raucao:
> Currently the numbers are not the same as on kredits-web. but I don't know why :)

Which numbers exactly?

bumi:
I ran this script and it gives me: 

```
Total confirmed Kredits: 1771000 between block 1 and 7765007
┌────┬───────────────┬─────────┐
│ ID │ Name          │ Kredits │
├────┼───────────────┼─────────┤
│ 1  │ Bumi          │ 260500  │
├────┼───────────────┼─────────┤
│ 2  │ Râu Cao       │ 421000  │
├────┼───────────────┼─────────┤
│ 3  │ Manuel        │ 204500  │
├────┼───────────────┼─────────┤
│ 4  │ galfert       │ 278500  │
├────┼───────────────┼─────────┤
│ 5  │ Greg          │ 336500  │
├────┼───────────────┼─────────┤
│ 6  │ Nick          │ 252000  │
├────┼───────────────┼─────────┤
│ 7  │ haythem96     │ 13000   │
...
```

which is different from what I see on kredits-web

raucao:
That's either a logical bug in the contract function for receiving overall kredits per contributor, or a bug in the script.

galfert:
When I use the filter in Kredits Web to search for all contributions of "haythem96", all contributions amount to 13000 Kredits. That's the same amount as @bumi got with the script. But Kredits Web shows 13500 in the toplist.

raucao:
> When I use the filter in Kredits Web to search for all contributions of "haythem96", all contributions amount to 13000 Kredits. That's the same amount as @bumi got with the script. But Kredits Web shows 13500 in the toplist.

That would point to a bug in the contract function then. The toplist is not calculated from the contributions directly, but using a contract function to receive those numbers. My hunch would be that vetoed contributions are counted (or not counted) in some place where they shouldn't be.

raucao:
How do I run the script against Rinkeby?

bumi:
I run `truffle exec --network=rinkeby scripts/list-contributions-per-contributor.js` (with local IPFS running)

raucao:
> That would point to a bug in the contract function then. The toplist is not calculated from the contributions directly, but using a contract function to receive those numbers. My hunch would be that vetoed contributions are counted (or not counted) in some place where they shouldn't be.

I had a quick look and the only thing I can see is that the if condition might ignore parentheses, but it looks correct to me. Maybe a bit too much for one line and such a critical piece of code, even if simple. Logic still dictates that it's a problem with that function, if the numbers don't add up otherwise.

raucao (approved):
We confirmed that the script itself is correct in the meantime.

PR #211, opened 2021-05-20
Add script for finding block closest to given date

raucao:
Useful for manual grant cycle management.


bumi (approved):
utACK. 
looks good to me.

PR #215, opened 2022-05-19
Load contract APIs from hardhat artifacts

bumi:
This finally removes all old/obsolete build files and scripts, and copies the correct ABIs to `lib` from the hardhat `artifacts` directory.


raucao:
In order to remove all old/obsolete files, and build everything from scratch, remove these directories first (some of them will also appear in your Git status now, since they've been removed from `.gitignore`):

    rm -rf apps artifacts build cache flattened_contracts

raucao (approved):
Works. :partying_face:

raucao:
The kredits label was missing here...

